PR #7, opened 2018-11-21
Added a simple instancing test; scene_graph makes the batches, render…

Nielsbishere:
…_system renders them. Updated shader to support instances and create render target now defaults to render target (not back buffer).

### DIFF
--- a/src/d3d12/d3d12_renderer.cpp
+++ b/src/d3d12/d3d12_renderer.cpp
@@ -36,10 +36,6 @@ namespace wr
 
 	void D3D12RenderSystem::Init(std::optional<Window*> window)
 	{
-		auto heapSize = 32;
-		auto heapAlignment = D3D12_DEFAULT_RESOURCE_PLACEMENT_ALIGNMENT;
-		auto x = (heapSize / heapAlignment + ((heapSize % heapAlignment) != 0 ? 1 : 0))*(heapAlignment);
-
 		m_device = d3d12::CreateDevice();
 		m_direct_queue = d3d12::CreateCommandQueue(m_device, d3d12::CmdListType::CMD_LIST_DIRECT);
 
@@ -56,11 +52,11 @@ namespace wr
 
 		// Temporary
 		// Create Constant Buffer Heap
-		constexpr auto model_cbs_size = SizeAlign(sizeof(temp::Model_CBData), 256);
-		constexpr auto cam_cbs_size = SizeAlign(sizeof(temp::ProjectionView_CBData), 256);
+		constexpr auto model_cbs_size = SizeAlign(sizeof(temp::Model_CBData), 256) * d3d12::settings::num_back_buffers;
+		constexpr auto cam_cbs_size = SizeAlign(sizeof(temp::ProjectionView_CBData), 256) * d3d12::settings::num_back_buffers;
 		constexpr auto sbo_size = 
-			model_cbs_size * 1 /* TODO: Make this more dynamic; right now it only supports 1 model */ 
-			+ cam_cbs_size * d3d12::settings::num_back_buffers;
+			(model_cbs_size * 2) /* TODO: Make this more dynamic; right now it only supports 2 mesh nodes */ 
+			+ cam_cbs_size;
 
 		m_cb_heap = d3d12::CreateHeap_SBO(m_device, sbo_size, d3d12::ResourceType::BUFFER, d3d12::settings::num_back_buffers);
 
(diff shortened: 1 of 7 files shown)


src/d3d12/d3d12_renderer.cpp
```diff
@@ -180,7 +176,7 @@ namespace wr
 		d3d12::Signal(m_fences[frame_idx], m_direct_queue);
 	}
 
-	void D3D12RenderSystem::RenderSceneGraph(SceneGraph const & scene_graph)
+	void D3D12RenderSystem::RenderSceneGraph(SceneGraph& scene_graph)
```
VZout: This should really be const. You wouldn't expect changes in a render func.

src/d3d12/d3d12_renderer.cpp
```diff
@@ -246,17 +234,52 @@ namespace wr
...
-		{
-			auto n_mesh = static_cast<D3D12Mesh*>(mesh);
-			auto frame_idx = GetFrameIdx();
+	}
+
+	//Render batches
+	void D3D12RenderSystem::RenderMeshBatches(SceneGraph& scene_graph)
+	{
+		
+		auto &batches = scene_graph.GetBatches();
```
VZout: I don't believe the style guide mentions this but please put the reference or pointer symbols to the left or centered. Everywhere else its either left or centered.

src/d3d12/d3d12_renderer.cpp
```diff
@@ -246,17 +234,52 @@ namespace wr
...
+			scene_graph.Optimize();
+
+		//Render batches
+		for (auto &elem : scene_graph.GetBatches()) {
+
+			Model *model = elem.first;
+			temp::MeshBatch &batch = elem.second;
+
+			//Update object data
+			d3d12::UpdateConstantBuffer(batch.batchBuffer->m_native, GetFrameIdx(), &batch.data, sizeof(batch.data));
```
VZout: Don't update constant buffers here. Should preferably happen somewhere that makes more sense (a place/time where also other buffers get updated).

src/d3d12/d3d12_renderer.hpp
```diff
@@ -3,16 +3,34 @@
 #include "../renderer.hpp"
 
 #include <DirectXMath.h>
+#include <unordered_map>
```
VZout: Unused?

src/d3d12/d3d12_renderer.hpp
```diff
@@ -3,16 +3,34 @@
...
 
 #include "../vertex.hpp"
 #include "d3d12_structs.hpp"
 
 namespace wr
 {
-
+	struct Model;
```
VZout: Unused?

src/scene_graph/mesh_node.hpp
```diff
@@ -1,10 +1,11 @@
 #pragma once
 
 #include "scene_graph.hpp"
+#include <unordered_map>
```
VZout: Unused?

src/scene_graph/scene_graph.cpp
```diff
@@ -52,4 +59,32 @@ namespace wr
 		return nullptr;
 	}
 
+	temp::MeshBatches& SceneGraph::GetBatches() { return m_batches; }
+
+	void SceneGraph::Optimize() {
+
```
VZout: This function contains some new lines at the beginning and end.

src/scene_graph/scene_graph.hpp
```diff
@@ -8,6 +8,8 @@
 #include "../util/defines.hpp"
 #include "../resource_pool.hpp"
 
+#define MAX_INSTANCE_COUNT 768
```
VZout: Prefer a `static const constexpr` in `d3d12_settings.hpp` over a definition

src/scene_graph/scene_graph.hpp
```diff
@@ -59,12 +87,18 @@ namespace wr
 		void RemoveChildren(std::shared_ptr<Node> const & parent);
 		std::shared_ptr<CameraNode> GetActiveCamera();
 
+		void Optimize();
+		temp::MeshBatches &GetBatches();
```
VZout: Getter should be const.

src/scene_graph/scene_graph.hpp
```diff
@@ -40,6 +42,32 @@ namespace wr
...
+		};
+
+		struct MeshBatch
+		{
+			unsigned int num_instances = 0;
+			D3D12ConstantBufferHandle* batchBuffer;
+			Model_CBData data;
+		};
+
+		typedef std::unordered_map<Model*, MeshBatch> MeshBatches;
```
VZout: Prefer `using` over `typedef`

src/d3d12/d3d12_renderer.cpp
Nielsbishere: Not possible; SceneGraph::Optimize sets up the batches inside of RenderSceneGraph -> RenderMeshBatches; which requires write access.

src/d3d12/d3d12_renderer.cpp
Nielsbishere: Moved to optimize

src/scene_graph/scene_graph.hpp
Nielsbishere: That would require scene_graph.hpp to include d3d12_settings.hpp though

PR #8, opened 2018-11-21
First itteration of registries.

VZout:
Added root signature, shader and pipeline registries.


src/d3d12/d3d12_root_signature.cpp
```diff
@@ -14,7 +14,7 @@ namespace wr::d3d12
 		return root_signature;
 	}
 
-	void FinalizeRootSignature(RootSignature* root_signature, Device* device)
+	void FinalizeRootSignature(d3d12::RootSignature* root_signature, Device* device)
```
VZout: Debug change

src/renderer.hpp
```diff
@@ -16,6 +18,7 @@ namespace wr
 	class SceneGraph;
 	class MaterialPool;
 	class ModelPool;
+	class PipelineRegistry;
```
VZout: unused

PR #29, opened 2018-11-27
Feature lighting

Nielsbishere:
Added LightNode for directional, spot and point lights


src/util/log.hpp
```diff
@@ -5,6 +5,7 @@
 #define LOG_CALLBACK
```
VZout: I'm not a fan of this design change. I agree that this functionality would be handy but preferably I'd use a preprocessor definition instead of a boolean.

src/render_tasks/d3d12_imgui_render_task.hpp
```diff
@@ -51,8 +51,8 @@ namespace wr
 			ImGui_ImplDX12_Init(n_render_system.m_device->m_native,
 				d3d12::settings::num_back_buffers,
 				(DXGI_FORMAT)d3d12::settings::back_buffer_format,
-				d3d12::GetCPUHandle(data.out_descriptor_heap).m_native,
-				d3d12::GetGPUHandle(data.out_descriptor_heap).m_native);
+				d3d12::GetCPUHandle(data.out_descriptor_heap, 0 /* TODO: Version here? */).m_native,
```
VZout: Yeah, using the version their seems fine to me.

src/render_tasks/d3d12_deferred_composition.hpp
```diff
@@ -34,16 +34,23 @@ namespace wr
...
+			for (uint32_t i = 0; i < d3d12::settings::num_back_buffers; ++i)
+			{
+
+				auto cpu_handle = d3d12::GetCPUHandle(data.out_srv_heap, i);
+
+				auto deferred_main_data = fg->GetData<DeferredMainTaskData>();
+				auto deferred_main_rt = data.out_deferred_main_rt = static_cast<D3D12RenderTarget*>(deferred_main_data.m_render_target);
+				d3d12::CreateSRVFromRTV(deferred_main_rt, cpu_handle, 2, deferred_main_data.m_rt_properties.m_rtv_formats.data());
+				d3d12::CreateSRVFromDSV(deferred_main_rt, cpu_handle);
+				d3d12::CreateSRVFromStructuredBuffer(n_render_system.GetLightBuffer(), cpu_handle, i);
```
VZout: The scene graph should return the light buffer. The render system has no idea what scene graph is being rendered. We want to be able to switch scene graphs.

src/d3d12/d3d12_renderer.hpp
```diff
@@ -68,6 +69,8 @@ namespace wr
 		wr::CommandList* GetComputeCommandList(unsigned int num_allocators) final;
 		wr::CommandList* GetCopyCommandList(unsigned int num_allocators) final;
 		RenderTarget* GetRenderTarget(RenderTargetProperties properties) final;
+		d3d12::HeapResource* GetLightBuffer();
```
VZout: Should be part of the scene graph.

src/d3d12/d3d12_renderer.hpp
```diff
@@ -96,10 +101,14 @@ namespace wr
 
 		// temporary
 		d3d12::Heap<HeapOptimization::SMALL_BUFFERS>* m_cb_heap;
+		d3d12::Heap<HeapOptimization::BIG_STATIC_BUFFERS>* m_sb_heap;
```
VZout: You should now have access to koen's platform independent pool's so you can put this in the scene graph instead.

src/d3d12/d3d12_renderer.hpp
```diff
@@ -96,10 +101,14 @@ namespace wr
...
 
 		d3d12::Viewport m_viewport;
 		d3d12::CommandList* m_direct_cmd_list;
 		d3d12::StagingBuffer* m_fullscreen_quad_vb;
+		d3d12::HeapResource* m_light_buffer;
+
+		Light* m_lights;
```
VZout: Please don't use C style arrays, prefer std::array and std::vector. C style arrays are relatively error prone.

src/d3d12/d3d12_renderer.cpp
```diff
@@ -416,6 +443,24 @@ namespace wr
 		}
 	}
 
+	void D3D12RenderSystem::Update_LightNodes(std::vector<std::shared_ptr<LightNode>>& nodes, CommandList* cmd_list)
+	{
+		auto n_cmd_list = static_cast<D3D12CommandList*>(cmd_list);
+
+		uint32_t count = 0;
+
+		for (auto& node : nodes)
```
VZout: prefer a (non-range based) for loop if you require the index.

src/d3d12/d3d12_heap.cpp
```diff
@@ -466,13 +466,14 @@ namespace wr::d3d12
...
 
-	HeapResource * AllocStructuredBuffer(Heap<HeapOptimization::BIG_STATIC_BUFFERS>* heap, std::uint64_t size_in_bytes, std::uint64_t stride)
+	HeapResource * AllocStructuredBuffer(Heap<HeapOptimization::BIG_STATIC_BUFFERS>* heap, std::uint64_t size_in_bytes, std::uint64_t stride, bool used_as_uav)
 	{
 		auto cb = new HeapResource();
 		decltype(Device::m_native) n_device;
 		heap->m_native->GetDevice(IID_PPV_ARGS(&n_device));
 		cb->m_unaligned_size = size_in_bytes;
 		cb->m_stride = stride;
+		cb->m_used_as_uav = used_as_uav;
```
VZout: Is this member variable necessary? I don't see it used anywhere beside this function.

src/d3d12/d3d12_descriptor_heap.cpp
```diff
@@ -14,22 +14,31 @@ namespace wr::d3d12
...
-		HRESULT hr = n_device->CreateDescriptorHeap(&heap_desc, IID_PPV_ARGS(&heap->m_native));
+		for (uint32_t i = 0; i < descriptor.m_versions; ++i)
+		{
+
+			D3D12_DESCRIPTOR_HEAP_DESC heap_desc;
+			heap_desc.NumDescriptors = descriptor.m_num_descriptors;
+			heap_desc.Type = (D3D12_DESCRIPTOR_HEAP_TYPE)descriptor.m_type;
+			heap_desc.Flags = descriptor.m_shader_visible ? D3D12_DESCRIPTOR_HEAP_FLAG_SHADER_VISIBLE : D3D12_DESCRIPTOR_HEAP_FLAG_NONE;
+			heap_desc.NodeMask = 0;
+			HRESULT hr = n_device->CreateDescriptorHeap(&heap_desc, IID_PPV_ARGS(&heap->m_native[i]));
```
VZout: We have a macro that does the HRESULT for you and logs a message when it fails.. Its called `TRY_M` `TRY_M(result, "mymessage")`

deferred_composition.hlsl
```diff
@@ -27,6 +45,52 @@ float3 world_to_view(float4 pos, float4x4 view) {
...
+	float3 light_color = light.col;
+
+	float3 radiance = light_color * 3;
+	float3 lighting = BRDF(L, V, normal, 0.4f, 0.6f, albedo, radiance, light_color) * intensity;
+
+	return lighting;
+
+}
+
+float3 shade(float3 vpos, float3 V, float3 albedo, float3 normal)
```
VZout: I don't like having 2 functions called shade.

src/d3d12/d3d12_heap.cpp
Nielsbishere: m_used_as_uav is needed to transition it properly after updating. The older implementation of UpdateStructuredBuffer assumed UAV, while that is almost never the case.

src/render_tasks/d3d12_imgui_render_task.hpp
Nielsbishere: Would require the out_descriptor_heap for imgui to be versioned

src/render_tasks/d3d12_deferred_composition.hpp
Nielsbishere: Not available yet - but, Koen is now working on a StructuredBuffer resource pool and it will be commited into this branch

src/d3d12/d3d12_renderer.hpp
Nielsbishere: Not available yet - but, Koen is now working on a StructuredBuffer resource pool and it will be commited into this branch

PR #36, opened 2018-11-28
Feature resize

VZout:
I implemented resizing support and fixed a bug with signalling.

+ Swapchain now sets minimum frame latency.
+ Resizing is now properly implemented.
+ Fullscreen works as expected now.
+ The ability to request fullscreen mode. (Recommended over alt+enter)
+ Signal is no longer being called twice after every execute.
+ You can now (`d3d12::ResizeViewport`) resize the viewport.

### DIFF
--- a/src/d3d12/d3d12_settings.hpp
+++ b/src/d3d12/d3d12_settings.hpp
@@ -29,7 +29,7 @@ namespace wr::d3d12::settings
 	static const constexpr DXGI_ALPHA_MODE swapchain_alpha_mode = DXGI_ALPHA_MODE_UNSPECIFIED;
 	static const constexpr bool enable_gpu_timeout = false;
 	static const constexpr bool enable_debug_factory = true;
-	static const constexpr DebugLayer enable_debug_layer = DebugLayer::DISABLE;
+	static const constexpr DebugLayer enable_debug_layer = DebugLayer::ENABLE_WITH_GPU_VALIDATION;
 	static const constexpr char* default_shader_model = "5.0";
 	static const constexpr std::uint8_t num_back_buffers = 3;
 	static const constexpr std::uint32_t num_instances_per_batch = 768U;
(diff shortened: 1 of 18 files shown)


src/window.hpp
```diff
@@ -55,6 +57,7 @@ namespace wr
 		LRESULT CALLBACK WindowProc_Impl(HWND, UINT, WPARAM, LPARAM);
 
 		KeyCallback m_key_callback;
+	public:
```
VZout: This was for debugging.

src/window.hpp
```diff
@@ -47,6 +47,8 @@ namespace wr
 		std::int32_t GetHeight() const;
 		/*! Returns the native window handle (HWND)*/
 		HWND GetWindowHandle() const;
+		/*! Checks whether the window is fullscreen */
+		bool IsFullscreen() const;
```
VZout: Should be removed. This is quite a poor way to do this. No good alternative at the moment so I'll leave this as is.

src/window.cpp
```diff
@@ -138,6 +138,17 @@ namespace wr
 		return m_handle;
 	}
 
+	bool Window::IsFullscreen() const
+	{
+		RECT a, b;
```
VZout: Needs rework.

src/d3d12/d3d12_settings.hpp
```diff
@@ -29,7 +29,7 @@ namespace wr::d3d12::settings
 	static const constexpr DXGI_ALPHA_MODE swapchain_alpha_mode = DXGI_ALPHA_MODE_UNSPECIFIED;
 	static const constexpr bool enable_gpu_timeout = false;
 	static const constexpr bool enable_debug_factory = true;
-	static const constexpr DebugLayer enable_debug_layer = DebugLayer::ENABLE_WITH_GPU_VALIDATION;
+	static const constexpr DebugLayer enable_debug_layer = DebugLayer::DISABLE;
```
VZout: Was for debugging purposes. Shouldn't be commited.

PR #38, opened 2018-11-29
Feature lighting

Nielsbishere:
Using platform independent constant & structured buffers


src/scene_graph/scene_graph.hpp
```diff
@@ -92,11 +90,11 @@ namespace wr
 		// Impl Functions
 		static std::function<void(RenderSystem*, temp::MeshBatches&, CommandList*)> m_render_meshes_func_impl;
 		static std::function<void(RenderSystem*, std::vector<std::shared_ptr<MeshNode>>&)> m_init_meshes_func_impl;
-		static std::function<void(RenderSystem*, std::vector<std::shared_ptr<CameraNode>>&)> m_init_cameras_func_impl;
+		static std::function<void(RenderSystem*, std::vector<std::shared_ptr<CameraNode>>&, ConstantBufferPool*)> m_init_cameras_func_impl;
```
VZout: Constant buffer pool shouldn't be passed here. A camera node might use a different pool than another camera node. (Would be bad but we need to support it.) Instead have a reference/handle to the correct pool inside of the nood.

src/tests/HeapTests.cpp
```diff
@@ -0,0 +1,462 @@
+#include "../d3d12/d3d12_structs.hpp"
```
VZout: File name style guide.

src/tests/ModelResourcePoolTests.cpp
```diff
@@ -0,0 +1,144 @@
+#include "../d3d12/d3d12_structs.hpp"
```
VZout: File name style.

PR #46, opened 2018-11-30
Feature optimization

Nielsbishere:
Added several optimizations to VBO & RenderTarget that reduce bytes per pixel from 36 to 18 (50%) and bytes per vertex from 56 to 28 (50%).


VZout:
Really good work. But I can't accept this at the moment.
We want to do high level optimization before low level optimization.
This optimization increases the codes complexity in a way that is currently unnessessary.
I'll leave the branch alive so we can merge it during block B, C or D.

PR #69, opened 2018-12-06
Feature voidptr bridging

florianschut:


src/renderer.hpp
```diff
@@ -45,15 +45,15 @@ namespace wr
 
 		virtual void WaitForAllPreviousWork() = 0;
 
-		virtual CommandList* GetDirectCommandList(unsigned int num_allocators) = 0;
-		virtual CommandList* GetBundleCommandList(unsigned int num_allocators) = 0;
-		virtual CommandList* GetComputeCommandList(unsigned int num_allocators) = 0;
-		virtual CommandList* GetCopyCommandList(unsigned int num_allocators) = 0;
-		virtual RenderTarget* GetRenderTarget(RenderTargetProperties properties) = 0;
-		virtual void ResizeRenderTarget(RenderTarget** render_target, std::uint32_t width, std::uint32_t height) = 0;
+		virtual CommandList GetDirectCommandList(unsigned int num_allocators) = 0;
```
LaisoEmilio: This is a perfect example of why typedefing CommandList and RenderTarget to void* can cause confusion in reading the code. It was our mistake. It's better to typedef CommandList to void so that we can use CommandList* as a return type in example and make sure the users see that it's a pointer

src/d3d12/d3d12_renderer.cpp
```diff
@@ -332,9 +332,9 @@ namespace wr
 
 	void D3D12RenderSystem::StopRenderTask(CommandList* cmd_list, std::pair<RenderTarget*, RenderTargetProperties> render_target)
 	{
-		auto n_cmd_list = static_cast<D3D12CommandList*>(cmd_list);
-		auto n_render_target = static_cast<D3D12RenderTarget*>(render_target.first);
-		auto frame_idx = GetFrameIdx();
+		auto n_cmd_list = static_cast<d3d12::CommandList*>(cmd_list);
+		auto n_render_target = static_cast<d3d12::RenderTarget*>(render_target.first);
+		unsigned int frame_idx = GetFrameIdx();
```
VZout: Why did you change this?

florianschut: My bad, changed that when I had some unexplainable errors which where suppressed by specifically specifying a type. Later it turned out to be caused by a different error. I’ll change this back in my next submit!

PR #80, opened 2018-12-11
Feature named resources

florianschut:


src/root_signature_registry.hpp
```diff
@@ -21,6 +21,7 @@ namespace wr
 		std::vector<d3d12::desc::SamplerDesc> m_samplers; // TODO: Move to platform independed location
 		bool m_rtx = false;
 		bool m_rtx_local = false;
+		std::wstring name = L"";
```
VZout: Default name shouldn't be empty. Call it a "Unkown [type]" or something.

PR #81, opened 2018-12-11
Feature fixing issue 61

florianschut:


src/d3d12/d3d12_device.cpp
```diff
@@ -13,13 +13,13 @@ namespace wr::d3d12
 	{
 		void EnableDebugLayer(Device* device)
 		{
-			if (!(settings::enable_debug_layer & settings::DebugLayer::DISABLE)) // If the debug layer isn't disabled
+			if (!(settings::enable_debug_layer == settings::DebugLayer::DISABLE)) // If the debug layer isn't disabled
```
Nielsbishere: There's a != operator for that

src/d3d12/d3d12_device.cpp
```diff
@@ -13,7 +13,7 @@ namespace wr::d3d12
 	{
 		void EnableDebugLayer(Device* device)
 		{
-			if (!(settings::enable_debug_layer == settings::DebugLayer::DISABLE)) // If the debug layer isn't disabled
+			if (!(settings::enable_debug_layer != settings::DebugLayer::DISABLE)) // If the debug layer isn't disabled
```
Nielsbishere: That's not really what I meant though, you should probably remove that !()

PR #86, opened 2018-12-13
Feature materials and textures

LaisoEmilio:
Added materials in the mesh node rendering function. Right now the materials and textures are manually loaded and initialized in the demo's resources.hpp. In the future, the model loader should handle all of this with the help of the texture and material pools.
Right now the material only uses albedo and normal for rendering, but adding the rest of the PBR textures should not be a problem.


src/d3d12/d3d12_renderer.cpp
```diff
@@ -796,9 +814,62 @@ namespace wr
...
 							static_cast<D3D12ModelPool*>(model->m_model_pool)->GetIndexStagingBuffer()->m_size);
+
 						m_bound_model_pool = static_cast<D3D12ModelPool*>(model->m_model_pool);
 						m_bound_model_pool_stride = n_mesh->m_vertex_staging_buffer_stride;
 					}
+
+
+					enum MaterialPBR: unsigned int
+					{
+						PBR_Albedo = 0,
```
VZout: Enums should be all capitals https://teamwisp.github.io/code_standards/

VZout: Also a bit of a weird place to define this enum.

src/d3d12/d3d12_renderer.cpp
```diff
@@ -796,9 +814,62 @@ namespace wr
...
+						PBR_Normal,
+
+						PBR_Count
+					};
+					
+					d3d12::BindDescriptorHeaps(n_cmd_list, { m_rendering_heap }, frame_idx);
+
+					auto material_handle = mesh.second;
+					
+					if (material_handle != m_last_material)
```
VZout: Abstract all this code to a different function. This render function is becoming really fucking bad.

src/material_pool.hpp
```diff
@@ -15,6 +15,54 @@ struct aiMaterial;
...
+		TextureHandle& RoughnessMetallic() { return m_rough_metallic; }
+		void SetRoughMetallic(TextureHandle rough_met);
+
+		TextureHandle& AmbientOcclusion() { return m_rough_metallic; }
+		void SetAmbientOcclusion(TextureHandle ao);
+
+		bool IsAlphaMasked() { return m_alpha_masked; }
+		void SetAlphaMasked(bool alpha_masked) { m_alpha_masked = alpha_masked; }
+
+		bool IsDoubleSided() { return m_double_sided; }
```
VZout: Currently probably unused. So comment them out or mark them as deprecated. Having this code available will just give the impression we support it.

src/d3d12/d3d12_renderer.cpp
LaisoEmilio: Done.

src/material_pool.hpp
LaisoEmilio: Done 👍

PR #87, opened 2018-12-13
Feature rtx

VZout:
Basic unpolished rtx impl. Requires a lot of changes but should be committed now so we can work on different rt features.


VZout:
Merging so we can work on other RT features more easily.

PR #88, opened 2018-12-13
Feature model loading

161563:
Abstracted model loader and switched model pool over to use new model loader class. Also created an assimp implementation of the model loader class.


demo/resources.hpp
```diff
@@ -122,7 +122,7 @@ namespace resources
 		}
 
 		{
-			test_model = model_pool->Load<wr::Vertex>(material_pool.get(), texture_pool.get(), "resources/models/xbot.fbx", wr::ModelType::FBX);
+			test_model = model_pool->Load<wr::Vertex>(material_pool.get(), texture_pool.get(), "resources/models/xbot.fbx");
```
LaisoEmilio: 👌

src/model_loader.cpp
```diff
@@ -0,0 +1,111 @@
+#include "model_loader.hpp"
+
+namespace wr
+{
+	ModelLoader::~ModelLoader()
+	{
+		for (ModelData* model : m_loaded_models)
```
LaisoEmilio: Consider using the DeleteModel function instead of duplicating code. There are ways to safely loop through a vector while deleting elements in it, but if that requires too much time and you have other tasks I can still accept this.

src/model_loader.cpp
```diff
@@ -0,0 +1,111 @@
...
+		for (int i = 0; i < model->m_skeleton_data->m_bones.size(); ++i)
+		{
+			delete model->m_skeleton_data->m_bones[i];
+		}
+
+		delete model->m_skeleton_data;
+		delete model;
+	}
+
+	std::vector<std::string> ModelLoader::SupportedModelFormats()
```
LaisoEmilio: Is there any way you can get rid of the vector of strings while achieving the same result? An enum maybe? 

Look at this, different example, but can something like that be done for the supported model formats as well?
https://github.com/jpvanoosten/LearningDirectX12/blob/master/DX12Lib/src/Texture.cpp#L258

src/model_loader.cpp
```diff
@@ -0,0 +1,111 @@
...
+	{
+		return m_supported_model_formats;
+	}
+
+	bool ModelLoader::SupportsModelFormat(std::string model_type)
```
LaisoEmilio: if strings cannot be avoided, can this be a const reference? does it need to be a string or can it be a string_view?

src/model_loader.hpp
```diff
@@ -0,0 +1,187 @@
...
+	struct EmbeddedTexture;
+
+	enum class TextureLocation
+	{
+		EXTERNAL = 0,
+		EMBEDDED,
+		NON_EXISTENT
+	};
+
+	struct Vector3
```
LaisoEmilio: Why the need for custom Vector3/4 and Matrix classes? And if they can't be removed, they should at least be hidden. Putting them in the wr namespace and in an hpp file means that you're exposing them to the user and could be used outside your original intentions.

src/model_loader_assimp.cpp
```diff
@@ -0,0 +1,348 @@
+#include "model_loader_assimp.hpp"
+
+#include "util/log.hpp"
+
+namespace wr
+{
+
+	AssimpModelLoader::AssimpModelLoader()
+	{
+		m_supported_model_formats.push_back("obj");
```
LaisoEmilio: Initializer list could be better. More readable and probably faster (Viktor said)

src/model_loader_assimp.cpp
```diff
@@ -0,0 +1,348 @@
...
+		m_supported_model_formats.push_back("fbx");
+		m_supported_model_formats.push_back("raw");
```
LaisoEmilio: Also, now I see the reason for strings, so ignore the other comment on them.

src/model_pool.hpp
```diff
@@ -157,203 +147,180 @@ namespace wr
...
-			aiProcess_CalcTangentSpace |
-			aiProcess_JoinIdenticalVertices |
-			aiProcess_OptimizeMeshes |
-			aiProcess_ImproveCacheLocality |
-			aiProcess_MakeLeftHanded);
-
-		if (!scene || scene->mFlags & AI_SCENE_FLAGS_INCOMPLETE || !scene->mRootNode)
+		IS_PROPER_VERTEX_CLASS(TV);
+
+		AssimpModelLoader assimp_loader();
```
LaisoEmilio: Consider static initialization of model loaders.

src/model_loader.cpp
161563: The idea here is that I can simply feed in the file extension. This also makes it easier to create model loaders for new formats, since you won't have to go around modifying the enum and other stuff.

src/model_loader.hpp
161563: I need to access this data in the model pool, and I can't use the DX12 versions.

src/model_pool.hpp
161563: Done. You still have to create the actual class somewhere. I'm currently using demo.cpp for that.

PR #92, opened 2018-12-14
Feature dynamic frame graph

161563:
Made renderer clear the back buffer if the current frame graph is different than the previous one
Frame graphs now have internal uid, for distinguishing them
Created small test system in demo for demonstrating frame graph changing and creation (F2 switched between first two frame graphs, F3 creates a third frame graph and switches to it)


src/d3d12/d3d12_renderer.cpp
```diff
@@ -167,6 +169,37 @@ namespace wr
 
 		d3d12::Begin(m_direct_cmd_list, frame_idx);
 
+		if (frame_graph.GetUID() != m_buffer_frame_graph_uids[frame_idx])
```
VZout: Abstract this into a separate function please. It is cluttering render.

demo/demo.cpp
```diff
@@ -20,6 +20,8 @@ std::shared_ptr<wr::SceneGraph> scene_graph;
 
 std::shared_ptr<wr::TexturePool> texture_pool;
 
+int frame_graph = 1;
```
VZout: Please revert the demo.cpp. I tested it and it works. No real need to keep this in demo. Would fit better in a test.

src/d3d12/d3d12_renderer.cpp
```diff
@@ -150,6 +150,8 @@ namespace wr
 		// Execute
 		d3d12::End(m_direct_cmd_list);
 		d3d12::Execute(m_direct_queue, { m_direct_cmd_list }, m_fences[frame_idx]);
+
+		m_buffer_frame_graph_uids.resize(d3d12::settings::num_back_buffers);
```
VZout: Does this need to be a array? Can't it just be a single value?

161563: I'm using an array because each of the back buffers needs to be cleared, and you can't clear them all at the same time. This allows me to know if I still need to clear this one or not.

PR #101, opened 2018-12-18
Feature PBR

florianschut:
Passed metallicness and roughness to the BRDF shader function and added a few resources to test it out as well as a new scene.


src/scene_graph/camera_node.cpp
```diff
@@ -18,7 +18,7 @@ namespace wr
 		DirectX::XMVECTOR right = DirectX::XMVector3Normalize(m_transform.r[0]);
 
 		m_view = DirectX::XMMatrixLookAtRH(pos, DirectX::XMVectorAdd(forward, pos), up);
-		m_projection = DirectX::XMMatrixPerspectiveFovRH(m_fov, m_aspect_ratio, 0.6, m_frustum_far);
+		m_projection = DirectX::XMMatrixPerspectiveFovRH(m_fov, m_aspect_ratio, 0.001, m_frustum_far);
```
VZout: Revert this please. This was for debugging.

VZout: This should be reverted. it was for debugging.

PR #108, opened 2019-01-07
Feature dxr materials

VZout:
- Path Tracing
- Accumulation
- Depth of Field
- Better transformations.


resources/shaders/accumulation.hlsl
```diff
@@ -0,0 +1,32 @@
...
+SamplerState s0 : register(s0);
+
+cbuffer CameraProperties : register(b0)
+{
+	float4x4 view;
+	float4x4 inv_projection_view;
+	float3 camera_position;
+	float padding;
+
+	float dep_light_radius;
```
LaisoEmilio: You should remove those, as you said.

src/imgui_tools.cpp
```diff
@@ -104,6 +104,7 @@ namespace wr::imgui::window
 
 		if (open_hardware_info)
 		{
+			if (ImGui::Button("Clear")) render_system.clear_path = true;
```
LaisoEmilio: Missing scope after the if statement.

PR #112, opened 2019-01-08
Feature scene graph materials

161563:
Materials of a model are now stored in its scene graph node, instead of in the model itself.


src/scene_graph/scene_graph.cpp
```diff
@@ -197,31 +197,83 @@ namespace wr
 			for (unsigned int i = 0; i < m_mesh_nodes.size(); ++i) {
 
 				auto node = m_mesh_nodes[i];
-				auto it = m_batches.find(node->m_model);
```
LaisoEmilio: Please split this code  into functions where possible and add comments as the code is not immediate to understand

161563:
Incorporated feedback.

src/scene_graph/scene_graph.cpp
```diff
@@ -185,10 +185,77 @@ namespace wr
...
+		++offset;
+	}
+
+	bool SceneGraph::TryInsertingNode(std::shared_ptr<MeshNode> mesh_node)
+	{
+		constexpr uint32_t max_size = d3d12::settings::num_instances_per_batch;
+
+		constexpr auto model_size = sizeof(temp::ObjectData) * max_size;
+
+		std::pair<std::unordered_multimap<Model*, temp::MeshBatch>::iterator,
```
Nielsbishere: Prefer to use auto over long variable names

src/scene_graph/mesh_node.hpp
```diff
@@ -8,10 +8,109 @@ namespace wr
 	struct MeshNode : Node
 	{
 		Model* m_model;
+		std::vector<MaterialHandle> m_materials;
 
 		DirectX::XMVECTOR m_aabb[2];
 
-		MeshNode(Model* model) : m_model(model) { }
+		MeshNode(Model* model) : m_model(model)
```
Nielsbishere: Please add a source file for all of the implementations of MeshNode. So you don't have to include any files and can use predeclaration.

VZout:
Closing for now. Seems state.

PR #118, opened 2019-01-14
Feature render texture output

tntmeijs:
# Render texture output

<hr>

## About
I have been working on a way to read the GPU pixel data on the CPU. This can be used for a variety of things. I plan on using this to copy the framebuffer data to the Maya viewport.

<hr>

## Changes
- New render pass to copy resources to a CPU-visible buffer.
- Framegraph now holds an optional data structure that stores the readback buffer data.
- Instead of returning a unique pointer to a texture, the renderer now returns the CPU texture memory address, as well as the size of the memory block.
- Render passes take in a width and height to prepare for Maya interation (or just windowless rendering in general).


demo/demo_frame_graphs.hpp
```diff
@@ -23,17 +24,25 @@ namespace fg_manager
...
 		{
 			auto& fg = frame_graphs[(int)PrebuildFrameGraph::RAYTRACING];
 			fg = new wr::FrameGraph(3);
 
 			wr::AddBuildAccelerationStructuresTask(*fg);
 			wr::AddRaytracingTask(*fg);
 			wr::AddAccumulationTask(*fg);
+			
+			// Copy the scene render pixel data to a buffer that is CPU-accessible
+			wr::AddRenderTargetReadBackTask<wr::RaytracingData>(*fg, std::nullopt, std::nullopt);
```
VZout: We probably want to put this into a different frame graph. We don't want to read-back when running as standalone.

PR #119, opened 2019-01-15
Feature multithreaded staging

161563:
Changed the staging of the model pool and the texture pool to use multithreading for memory transfers and command list generation.


src/d3d12/d3d12_command_queue.cpp
```diff
@@ -33,9 +33,16 @@ namespace wr::d3d12
...
+		if (fence != nullptr)
+		{
+			fence->m_fence_value++;
+			Signal(fence, cmd_queue);
+		}
+	}
 
+	void WaitFor(CommandQueue * cmd_queue, Fence * fence)
+	{
+		cmd_queue->m_native->Wait(fence->m_native, fence->m_fence_value);
```
VZout: Why did you need this? `Fence::WaitFor` should do the job right? And it won't require knowledge about the queue.

src/d3d12/d3d12_model_pool.hpp
```diff
@@ -46,7 +52,9 @@ namespace wr
 		void Evict() final;
 		void MakeResident() final;
 
-		void StageMeshes(d3d12::CommandList* cmd_list);
+		void StageMeshes();
```
VZout: Return the generated command lists instead and mark it as [[nodiscard]]. The user shouldn't have to access the command list as a member variable. It would be less error prone if you just return a constant reference to it.

src/d3d12/d3d12_model_pool.hpp
```diff
@@ -46,7 +52,9 @@ namespace wr
...
+
+		void WaitForStaging();
```
VZout: If the render system executes the command lists it will wait for you there automatically since it gets staged at the beginning of the frame. We want to abstract all the gpu related code into the render system to make the renderer easy to use.

src/d3d12/d3d12_resource_pool_texture.cpp
```diff
@@ -33,10 +34,30 @@ namespace wr
 		m_mipmapping_cpu_handle = d3d12::GetCPUHandle(m_mipmapping_heap, 0);
 		m_mipmapping_gpu_handle = d3d12::GetGPUHandle(m_mipmapping_heap, 0);
 
+		m_thread_pool = m_render_system.GetThreadPool();
+
+		m_command_lists.resize(d3d12::settings::max_threads + 1);
```
VZout: Why max thread + 1? max_threads should never be 0 since that would mean you don't have a main thread.

src/d3d12/d3d12_resource_pool_texture.cpp
```diff
@@ -33,10 +34,30 @@ namespace wr
...
+
+		for (auto& list : m_command_lists)
+		{
+			list = d3d12::CreateCommandList(m_render_system.m_device, 1, wr::CmdListType::CMD_LIST_DIRECT);
```
VZout: Could you use a copy cmd list instead for staging?

src/d3d12/d3d12_resource_pool_texture.cpp
```diff
@@ -33,10 +34,30 @@ namespace wr
...
+		}
+
+		m_command_recording_futures.resize(d3d12::settings::max_threads + 1);
+
+		m_staging_fence = d3d12::CreateFence(m_render_system.m_device);
```
VZout: Fence would be unnessessary if the render system executes the command lists.

src/d3d12/d3d12_resource_pool_texture.cpp
```diff
@@ -90,6 +111,109 @@ namespace wr
...
+					m_command_lists[m_command_lists.size() - 1], unstaged_textures);
+
+			for (auto& future : m_command_recording_futures)
+			{
+				future.get();
+			}
+
+			MoveStagedTextures();
+
+			d3d12::Execute(m_render_system.m_direct_queue, m_command_lists, m_staging_fence);
```
VZout: The pool shouldn't execute the staging fence. Should happen from the render system since its recommended to call execute as little as possible.

src/d3d12/d3d12_settings.hpp
```diff
@@ -45,5 +45,6 @@ namespace wr::d3d12::settings
 	static const constexpr bool enable_object_culling = false;
 	static const constexpr unsigned int num_max_rt_materials = 10;
 	static const constexpr unsigned int fallback_ptrs_offset = 26;
-	
-} /* wr::d3d12::settings */
\ No newline at end of file
+	static const constexpr unsigned int max_threads = 4;
```
VZout: Use `settings` instead for the maximum amount of threads since its not related to DX12.

src/d3d12/d3d12_staging_buffer.cpp
```diff
@@ -55,6 +54,7 @@ namespace wr::d3d12
 	void UpdateStagingBuffer(StagingBuffer* buffer, void * data, std::uint64_t size, std::uint64_t offset)
 	{
 		memcpy(static_cast<std::uint8_t*>(buffer->m_data) + offset, data, size);
+		memcpy(buffer->m_cpu_address + offset, static_cast<std::uint8_t*>(buffer->m_data) + offset, size);
```
VZout: 2 copies? Haven't looked at what your copying but make sure its actually necessary.

src/d3d12/d3d12_structs.hpp
```diff
@@ -7,6 +7,7 @@
 #include <utility>
 #include <dxcapi.h>
 #include <array>
+#include <mutex>
```
VZout: This is unused.

src/d3d12/d3d12_model_pool.hpp
```diff
@@ -78,11 +86,26 @@ namespace wr
...
 		
-		std::queue<internal::MeshInternal*> m_mesh_stage_queue;
+		std::vector<std::queue<internal::MeshInternal*>> m_mesh_stage_queues;
+
+		std::uint32_t m_current_queue;
 
 		std::uint64_t m_vertex_buffer_size;
 		std::uint64_t m_index_buffer_size;
 
+		util::ThreadPool* m_thread_pool;
```
VZout: You added a thread pool to the render system and here. You probably only need 1

src/d3d12/d3d12_model_pool.hpp
```diff
@@ -78,11 +86,26 @@ namespace wr
...
+
+		std::vector<d3d12::CommandList*> m_command_lists;
+		std::vector<d3d12::CommandList*> m_transition_command_lists;
```
VZout: Unused?

src/d3d12/d3d12_model_pool.hpp
```diff
@@ -78,11 +86,26 @@ namespace wr
...
+
+		std::vector<std::future<void>> m_command_recording_futures;
+		std::vector<bool> m_command_recording_launched;
+
+		d3d12::Fence* m_staging_fence;
+		d3d12::Fence* m_transition_fence;
```
VZout: Unused?

src/d3d12/d3d12_model_pool.hpp
```diff
@@ -78,11 +86,26 @@ namespace wr
...
 		std::uint64_t m_vertex_buffer_size;
 		std::uint64_t m_index_buffer_size;
 
+		util::ThreadPool* m_thread_pool;
+
+		std::vector<d3d12::CommandList*> m_command_lists;
+		std::vector<d3d12::CommandList*> m_transition_command_lists;
+
+		std::vector<std::future<void>> m_command_recording_futures;
+		std::vector<bool> m_command_recording_launched;
```
VZout: Unused?

src/d3d12/d3d12_resource_pool_texture.cpp
```diff
@@ -90,6 +111,109 @@ namespace wr
...
+			auto itr = m_unstaged_textures.begin();
+
+			for (itr; itr != m_unstaged_textures.end(); ++itr)
+			{
+				unstaged_textures.push_back(static_cast<d3d12::TextureResource*>(itr->second));
+			}
+
+			for (int i = 0; i < d3d12::settings::max_threads; ++i) {
+
+				d3d12::CommandList* cmdlist = m_command_lists[i];
```
VZout: styleguide

src/d3d12/d3d12_resource_pool_texture.cpp
```diff
@@ -90,6 +111,109 @@ namespace wr
...
+					}
+
+					d3d12::End(cmdlist);
+				},
+					cmdlist, unstaged_textures);
+
+			}
+
+			m_command_recording_futures[m_command_recording_futures.size() - 1] =
+				m_thread_pool->Enqueue([this](
```
VZout: MipMapping shouldn't be threaded like this. Should also split across x amount of frames. Just remove this from the thread pool for now. GenerateMipmaps needs refactoring. After that we will add multithreading support to that.

src/d3d12/d3d12_resource_pool_texture.cpp
```diff
@@ -90,6 +111,109 @@ namespace wr
...
+			for (itr; itr != m_unstaged_textures.end(); ++itr)
+			{
+				unstaged_textures.push_back(static_cast<d3d12::TextureResource*>(itr->second));
+			}
+
+			for (int i = 0; i < d3d12::settings::max_threads; ++i) {
+
+				d3d12::CommandList* cmdlist = m_command_lists[i];
+
+				m_command_recording_futures[i] = m_thread_pool->Enqueue([i](
```
VZout: This code can be greatly simplified if you use the `DivideWork` function from the thread pool. As I mentioned when I gave you this task. See the frame graph for how it works.

src/d3d12/d3d12_command_queue.cpp
161563: This makes the queue wait on the fence, instead of the program. It allows you to synchronize multiple queues without having to halt the cpu.

src/d3d12/d3d12_command_queue.cpp
```diff
@@ -33,9 +33,16 @@ namespace wr::d3d12
 
 		cmd_queue->m_native->ExecuteCommandLists(native_lists.size(), native_lists.data());
 
-		fence->m_fence_value++;
-		Signal(fence, cmd_queue);
+		if (fence != nullptr)
```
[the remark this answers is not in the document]
161563: Otherwise it would crash the program if you don't supply a fence, and sometimes it isn't necessary to synchronize stuff because you're already doing it on the GPU, through things like barriers.

src/d3d12/d3d12_model_pool.hpp
161563: This is the thread pool from the render system. I'm storing the pointer for easier access.

src/d3d12/d3d12_model_pool.hpp
161563: I'm submitting the copy commands on the copy queue, but I can't do transitions there. I'm using two arrays of command lists for this reason.

src/d3d12/d3d12_model_pool.hpp
161563: Yes, fixing it.

src/d3d12/d3d12_resource_pool_texture.cpp
161563: I'm using a fifth command list for mipmap generation. Combining the mipmap generation into each of the command lists breaks stuff.

src/d3d12/d3d12_resource_pool_texture.cpp
161563: No clue. It seems to be doing transitions though, so that might break it. I'll try it.

src/d3d12/d3d12_resource_pool_texture.cpp
161563: That would have been nice to have some documentation on, since I've no clue on how DivideWork works, and I can't remember being told the frame graph has an example. Also, wasn't that code bugged? I think Emilio found some weird error in it.

src/d3d12/d3d12_staging_buffer.cpp
161563: One copies to the staging buffer cpu memory, and one to the upload heap buffer. I think we're allocating cpu memory because we originally used updateSubResources, which required it. We should probably get rid of that in general though, it would save a lot of memory.

src/d3d12/d3d12_structs.hpp
161563: Fixing it.

VZout:
Closing for now. Seems state.

PR #126, opened 2019-01-17
Feature memory optimizations

161563:
Optimized memory usage by freeing up unused memory and making resource pools use bytes instead of megabytes.

Bench-marking returns the following improvements:
    Looking at task manager:
        1343 MB - Old
        273 MB - Optimized
        1070 MB - Difference
    Looking at Visual Studio memory monitor:
        1843 MB - Old
        728 MB - Optimized
        1115 MB - Difference
Edit: Figures no longer work with updates to engine, although there should still be a reduction in memory usage.


demo/resources.hpp
```diff
@@ -44,8 +44,8 @@ namespace resources
 
 	void CreateResources(wr::RenderSystem* render_system)
 	{
-		texture_pool = render_system->CreateTexturePool(256, 55);
-		material_pool = render_system->CreateMaterialPool(256);
+		texture_pool = render_system->CreateTexturePool(16*1024*1024, 55);
```
Nielsbishere: Might be an idea to add _MB or _KB (same way as _deg); though we as a team might have to discuss that

VZout: This. I won't except 1024*1024

161563: Done.

PR #130, opened 2019-01-18
Feature rt shadow

MeineZ:
**Description**
Implemented a new render pass that renders the scene in a hybrid pipeline (deferred -> ray tracing -> output). There has been added a new ray tracing shader, state object, root signature, shader library, and hybrid render task.

**Issues**
Multi-threading had to be disabled (src/settings) and `batch.num_instances` (src/d3d12/d3d12_renderer.cpp) in commented as these lines of code crashed the build of the acceleration structure. This causes the deferred renderer to stall (even jitter) all geometry that is supposed to be in motion (camera still works).

**Possible conflicts**
When adding new things (like a new descriptor range) to the full ray-tracing task, there is a chance that you also need to modify the same things in the hybrid pipeline.


demo/demo_frame_graphs.hpp
```diff
@@ -17,10 +18,11 @@ namespace fg_manager
 	{
 		RAYTRACING = 0,
 		DEFERRED = 1,
+		RT_HYBRID = 2,
 	};
 
-	static PrebuildFrameGraph current = fg_manager::PrebuildFrameGraph::DEFERRED;
-	static std::array<wr::FrameGraph*, 2> frame_graphs = {};
+	static PrebuildFrameGraph current = fg_manager::PrebuildFrameGraph::RT_HYBRID;
```
VZout: Change this back to deferred please :)

demo/demo.cpp
```diff
@@ -8,6 +8,7 @@
 
 #include "engine_interface.hpp"
 #include "scene_viknell.hpp"
+//#include "scene_rt_shadow.hpp"
```
VZout: Remove comments please.

demo/debug_camera.hpp
```diff
@@ -83,6 +83,17 @@ class DebugCamera : public wr::CameraNode
 		}
 	}
 
+	const float m_scroll_speed = 0.25f;
+
+	void MouseWheel(int amount)
+	{
+		float percent = (float) GET_WHEEL_DELTA_WPARAM(amount) / WHEEL_DELTA * m_scroll_speed;
```
VZout: Check if you hold rmb because other wise scrolling in imgui will scroll in the demo.

MeineZ:
The code in this pull request is already in the feature_rt_reflection pull request.

PR #133, opened 2019-01-18
Feature rt reflection

Nielsbishere:
# Description
Implemented a new render pass that renders the scene in a hybrid pipeline (deferred -> ray tracing -> output). There has been added a new ray tracing shader, state object, root signature, shader library, and hybrid render task.

# Possible conflicts
When adding new things (like a new descriptor range) to the full ray-tracing task, there is a chance that you also need to modify the same things in the hybrid pipeline.


src/d3d12/d3d12_root_signature.cpp
```diff
@@ -122,7 +122,8 @@ namespace wr::d3d12
 			HRESULT hr = D3D12SerializeRootSignature(&root_signature_desc, D3D_ROOT_SIGNATURE_VERSION_1, &signature, &error); //TODO: FIX error parameter
 			if (FAILED(hr))
 			{
-				LOGC("Failed to serialize root signature. Error: \n {}", (char*)error->GetBufferPointer());
+				char * err = (char*)error->GetBufferPointer();
```
VZout: Debug code?

PR #148, opened 2019-02-14
Fix lights again

Nielsbishere:
Fixed dynamically adding/removing lights


src/scene_graph/scene_graph.hpp
```diff
@@ -232,6 +234,8 @@ namespace wr
 			}
 		}
 
+		m_next_light_id = (uint32_t) m_light_nodes.size();
```
tntmeijs: C-style cast.

PR #149, opened 2019-02-14
[EDIT]: Fixed misalignment of sky boxes

florianschut:
Changed the deferred renderer to use equirectangular skybox to be more in line with (hybrid) raytracing. Made both raytracing pipelines use the PBR-util function to get the uvs for sampling the skybox so it works :)


tntmeijs:
Code looks fine to me. I'll trust you that it runs and works fine.
You'll need to ask @EmilioLaiso or @VZout for final approval, though.

PR #152, opened 2019-02-14
[EDIT]: Made all frame graphs resize on window callback

florianschut:


demo/demo_frame_graphs.hpp
```diff
@@ -97,6 +97,14 @@ namespace fg_manager
 		}
 	}
 
+	void Resize(wr::RenderSystem& render_system, std::uint32_t width, std::uint32_t height)
+	{
+		for (int i = 0; i < 3; ++i)
```
MeineZ: Don't think we want '3' to be hardcoded

florianschut: Fixed 💋

demo/demo_frame_graphs.hpp
```diff
@@ -97,6 +97,14 @@ namespace fg_manager
 		}
 	}
 
+	void Resize(wr::RenderSystem& render_system, std::uint32_t width, std::uint32_t height)
+	{
+		for (int i = 0; i < frame_graphs.size(); ++i)
```
VZout: could have used a range based loop.

PR #153, opened 2019-02-15
Feature fix mipmapping

LaisoEmilio:
Refactored mipmapping, added sRGB textures mipmapping (no BGR textures yet, they seem to work when treated as RGB). Added new texture formats support checking in the engine and functions to check and/or modify formats. Removed pow when sampling albedo textures in the shader cause they're now correctly sRGB.

### DIFF
--- a/src/d3d12/d3d12_device.cpp
+++ b/src/d3d12/d3d12_device.cpp
@@ -125,6 +125,59 @@ namespace wr::d3d12
 			device->m_adapter = (IDXGIAdapter4*)adapter;
 		}
 
+		void QueryForOptionalFormats(Device* device)
+		{
+			D3D12_FEATURE_DATA_D3D12_OPTIONS feature_data;
+			ZeroMemory(&feature_data, sizeof(feature_data));
+			HRESULT hr = device->m_native->CheckFeatureSupport(D3D12_FEATURE_D3D12_OPTIONS, &feature_data, sizeof(feature_data));
+
+			if (SUCCEEDED(hr))
+			{
+				// TypedUAVLoadAdditionalFormats contains a Boolean that tells you whether the feature is supported or not
+				if (feature_data.TypedUAVLoadAdditionalFormats)
+				{
+					// Can assume “all-or-nothing” subset is supported (e.g. R32G32B32A32_FLOAT)
+					// Cannot assume other formats are supported, so we check:
+
+					auto CheckAndSetFormat = [](Device* device, DXGI_FORMAT format)
+					{
+						D3D12_FEATURE_DATA_FORMAT_SUPPORT format_support = { format, D3D12_FORMAT_SUPPORT1_NONE, D3D12_FORMAT_SUPPORT2_NONE };
+						HRESULT hr = device->m_native->CheckFeatureSupport(D3D12_FEATURE_FORMAT_SUPPORT, &format_support, sizeof(format_support));
+						if (SUCCEEDED(hr) && (format_support.Support2 & D3D12_FORMAT_SUPPORT2_UAV_TYPED_LOAD) != 0)
+						{
+							device->m_optional_formats.set(format);
+						}
+					};
+
+					CheckAndSetFormat(device, DXGI_FORMAT_R16G16B16A16_UNORM);
+					CheckAndSetFormat(device, DXGI_FORMAT_R16G16B16A16_SNORM);
+					CheckAndSetFormat(device, DXGI_FORMAT_R32G32_FLOAT);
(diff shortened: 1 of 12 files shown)


src/d3d12/d3d12_device.cpp
```diff
@@ -125,6 +125,59 @@ namespace wr::d3d12
...
+
+			if (SUCCEEDED(hr))
+			{
+				// TypedUAVLoadAdditionalFormats contains a Boolean that tells you whether the feature is supported or not
+				if (feature_data.TypedUAVLoadAdditionalFormats)
+				{
+					// Can assume �all-or-nothing� subset is supported (e.g. R32G32B32A32_FLOAT)
+					// Cannot assume other formats are supported, so we check:
+
+					auto CheckAndSetFormat = [](Device* device, DXGI_FORMAT format)
```
VZout: function would be preferable here. No real benefit of using a lambda in this case.

src/d3d12/d3d12_functions.hpp
```diff
@@ -100,14 +100,26 @@ namespace wr::d3d12
...
 	//void CreateUAVFromTexture(TextureResource* tex, DescHeapCPUHandle& handle, unsigned int mip_slice = 0, unsigned int array_slice = 0);
 	void SetShaderSRV(wr::d3d12::CommandList* cmd_list, uint32_t rootParameterIndex, uint32_t descriptorOffset, TextureResource* tex);
+	void SetShaderSRV(wr::d3d12::CommandList* cmd_list, uint32_t rootParameterIndex, uint32_t descriptorOffset, d3d12::DescHeapCPUHandle& handle);
 	void SetShaderUAV(wr::d3d12::CommandList* cmd_list, uint32_t rootParameterIndex, uint32_t descriptorOffset, TextureResource* tex);
+	void SetShaderUAV(wr::d3d12::CommandList* cmd_list, uint32_t rootParameterIndex, uint32_t descriptorOffset, d3d12::DescHeapCPUHandle& handle);
+	void CopyResource(wr::d3d12::CommandList* cmd_list, TextureResource* src_texture, TextureResource* dst_texture);
 	void Destroy(TextureResource* tex);
 
+	// Format test and support functions
+	bool CheckUAVCompatibility(Format format);
```
VZout: These functions (or atleast not all) don't need to be globally defined I guess. write them in a `internal` namespace instead in the file you need them.

src/d3d12/d3d12_textures.cpp
```diff
@@ -11,7 +11,16 @@ namespace wr::d3d12
 	{
 		D3D12_RESOURCE_FLAGS flags = D3D12_RESOURCE_FLAG_NONE;
 
-		if (allow_uav) { flags |= D3D12_RESOURCE_FLAG_ALLOW_UNORDERED_ACCESS; }
+		if (allow_uav) 
+		{ 
+			//if (!d3d12::CheckUAVCompatibility(description->m_texture_format))
```
VZout: Please remove comments.

VZout: (commented out code)

src/d3d12/d3d12_textures.cpp
```diff
@@ -102,21 +111,94 @@ namespace wr::d3d12
...
+		}
+		else
+		{
+			if (tex->m_depth > 1)
+			{
+				//Then it's a 3D texture
+				dimension = D3D12_SRV_DIMENSION_TEXTURE3D;
+				srv_desc.Texture3D.MipLevels = tex->m_mip_levels;
+			}
+			else
```
VZout: can't this be `else if`?

src/d3d12/d3d12_textures.cpp
```diff
@@ -102,21 +111,94 @@ namespace wr::d3d12
...
+
+		if (tex->m_is_cubemap)
+		{
+			dimension = D3D12_SRV_DIMENSION_TEXTURECUBE;
+			srv_desc.TextureCube.MipLevels = mip_levels;
+			srv_desc.TextureCube.MostDetailedMip = most_detailed_mip;
 		}
 		else
 		{
 			if (tex->m_depth > 1)
```
VZout: Can't this be `else if`?

src/d3d12/d3d12_functions.hpp
LaisoEmilio: except for 1, they're all used in multiple places, so for now i'm leaving them like that since they are just checking functions that can't cause any harm to the underlying engine.

src/d3d12/d3d12_textures.cpp
LaisoEmilio: nope, those tests need to be executed only if the if statement fails

PR #155, opened 2019-02-15
[CHANGE] Attempt to fix resources being deleted too early

LaisoEmilio:
Attempt to fix resources being deleted before the command list that was referencing them was finished executing on the command queue

### DIFF
--- a/src/d3d12/d3d12_resource_pool_texture.hpp
+++ b/src/d3d12/d3d12_resource_pool_texture.hpp
@@ -59,7 +59,7 @@ namespace wr
 		DescriptorAllocator* m_mipmapping_allocator;
 
 		//Track resources that are created in one frame and destroyed after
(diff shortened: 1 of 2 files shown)


src/d3d12/d3d12_resource_pool_texture.hpp
```diff
@@ -59,7 +59,7 @@ namespace wr
 		DescriptorAllocator* m_mipmapping_allocator;
 
 		//Track resources that are created in one frame and destroyed after
-		std::vector<d3d12::TextureResource*> m_temporary_textures;
+		std::vector<d3d12::TextureResource*> m_temporary_textures[d3d12::settings::num_back_buffers];
```
VZout: we not programming C we are programming C++. prefer std array

LaisoEmilio: done 👍

PR #157, opened 2019-02-15
Fix Coordinate System

MeineZ:
All frame graphs are rendered in the same correct (right-handed) coordinate system now.

### DIFF
--- a/demo/debug_camera.hpp
+++ b/demo/debug_camera.hpp
@@ -108,19 +108,19 @@ public:
 	{
 		if (action == WM_KEYDOWN)
 		{
-			if (key == 0x57) // W
+			if (key == 'W')
 			{
 				m_forward_axis += -1;
 			}
-			if (key == 0x53) // S
+			if (key == 'S')
 			{
 				m_forward_axis += 1;
 			}
-			if (key == 0x41) // A
+			if (key == 'A')
 			{
 				m_right_axis += -1;
 			}
-			if (key == 0x44) // D
+			if (key == 'D')
 			{
 				m_right_axis += 1;
 			}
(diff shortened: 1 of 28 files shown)


demo/debug_camera.hpp
```diff
@@ -110,63 +110,63 @@ class DebugCamera : public wr::CameraNode
...
-				m_forward_axis += 1;
+				m_forward_axis += -1;
 			}
 			if (key == 0x53) // S
 			{
-				m_forward_axis += -1;
+				m_forward_axis += 1;
 			}
-			if (key == 0x44) // A
+			if (key == 0x41) // A
```
tntmeijs: We should probably create an enum for the keys sometime soon...

resources/shaders/deferred_composition.hlsl
```diff
@@ -36,9 +36,9 @@ void main_cs(int3 dispatch_thread_id : SV_DispatchThreadID)
 {
 	float2 screen_size = float2(0.f, 0.f);
 	output.GetDimensions(screen_size.x, screen_size.y);
-	float2 uv = float2(dispatch_thread_id.x / screen_size.x, 1.f - (dispatch_thread_id.y / screen_size.y));
+	float2 uv = float2(dispatch_thread_id.x / screen_size.x, dispatch_thread_id.y / screen_size.y);
```
tntmeijs: Cool, this may mean I no longer have to flip the Maya render texture...

demo/debug_camera.hpp
Nielsbishere: This can be replaced by 'A' be sure to match case.

tntmeijs: Oh yeah, didn't think of that. 👍

PR #161, opened 2019-02-17
[EDIT]: First itteration of the requested light editor changes

florianschut:


src/imgui_tools.cpp
```diff
@@ -187,78 +187,31 @@ namespace wr::imgui::window
...
 					}
 					else
 					{
+						light_selected = false;
 						selected_light = nullptr;
 					}
 				}
+				
+				if (i == 0)
+					lights[i]->m_light->tid |= (uint32_t)lights.size() << 2;
```
Nielsbishere: |= is invalid in this case, it should either `(x & 3) | (lightCount << 2)` or not be touched at all. lights[i] should be signaled.

PR #166, opened 2019-02-19
[CHANGE] Added BGR textures mipmapping and refactored the generate mips.

LaisoEmilio:
I moved some functions in the command_list.cpp file to match the order of the hpp file, it shows wierd in github but I didn't actually change those functions


src/d3d12/d3d12_resource_pool_texture.cpp
```diff
@@ -478,61 +485,12 @@ namespace wr
...
-		}
-
-		d3d12::Transition(d3d12_cmd_list, textures, ResourceState::UNORDERED_ACCESS, ResourceState::PIXEL_SHADER_RESOURCE);
-	}
-
 	void D3D12TexturePool::GenerateMips_UAV(d3d12::TextureResource* texture, CommandList* cmd_list)
 	{
 		wr::d3d12::CommandList* d3d12_cmd_list = static_cast<wr::d3d12::CommandList*>(cmd_list);
 
+		d3d12::Transition(d3d12_cmd_list, texture, texture->m_current_state, ResourceState::UNORDERED_ACCESS);
```
VZout: Could transition all at once but fuck that.

PR #170, opened 2019-02-21
Feature material system improvement

161563:
[CHANGE] Added ability to load textures from RAM.
[CHANGE] Added constant values to materials for color, alpha, metallic value and roughness. 
[CHANGE] Added flags to materials indicating if they have textures, if constants should be used, and other general flags.


resources/shaders/rt_hybrid.hlsl
```diff
@@ -218,13 +229,27 @@ void ReflectionHit(inout ReflectionHitInfo payload, in MyAttributes attr)
...
-	float metal = g_textures[material.metalicness_id].SampleLevel(s0, uv, 0).x;
+	float3 albedo = lerp(g_textures[material.albedo_id].SampleLevel(s0, uv, 0).xyz,
+		material.data.albedo_alpha.xyz,
+		use_albedo_constant != 0 || has_albedo_texture == 0);
+	float roughness = lerp(max(0.05, g_textures[material.roughness_id].SampleLevel(s0, uv, 0).x),
+		material.data.metallic_roughness.w,
+		use_roughness_constant != 0 || has_roughness_texture == 0);
+	float metal = lerp(g_textures[material.metalicness_id].SampleLevel(s0, uv, 0).x,
+		length(material.data.metallic_roughness.xyz),
+		use_metallic_constant != 0 || has_metallic_texture == 0);
```
LaisoEmilio: Looks good, consider making the above code, as well as the flags checking code into a separate function that can be reused in all the rendering pipelines. The textures can be passed in as parameters.

resources/shaders/material_util.hlsl
```diff
@@ -0,0 +1,120 @@
...
+	float4 roughness = lerp(material_metallic.SampleLevel(s0, input.uv, 0).y, data.metallic_roughness.w, use_roughness_constant != 0 || has_roughness_texture == 0);
+	float4 metallic = lerp(material_metallic.SampleLevel(s0, input.uv, 0).z, length(data.metallic_roughness.xyz), use_metallic_constant != 0 || has_metallic_texture == 0);
+#else
+	float4 roughness = lerp(max(0.05f, material_roughness.Sample(s0, uv)), data.metallic_roughness.wwww, use_roughness_constant != 0 || has_roughness_texture == 0);
+	float4 metallic = lerp(material_metallic.Sample(s0, uv), data.metallic_roughness.xyzx, use_metallic_constant != 0 || has_metallic_texture == 0);
+#endif
+	float3 tex_normal = lerp(material_normal.Sample(s0, uv).rgb * 2.0 - float3(1.0, 1.0, 1.0), float3(0.0, 0.0, 1.0), use_normal_texture == 0);
+	float3 normal = normalize(mul(tex_normal, tbn));
+
+	output.albedo_roughness = float4(albedo.xyz, roughness.r);
```
LaisoEmilio: Viktor: These don't need to be single variable in the output data. For usability we might just want a float3 for albedo, and a float for roughness...the same goes for normals.
In the gbuffers they're together for optimization, but here it's not needed.

resources/shaders/material_util.hlsl
```diff
@@ -0,0 +1,120 @@
...
+	output.normal_metallic = float4(normal, metallic.r);
+	return output;
+}
+
+OutputMaterialData InterpretMaterialDataRT(MaterialData data,
```
LaisoEmilio: Why no tbn here? For consistency it either goes in here as well or both the functions don't use it and it's left in the shader code.

src/d3d12/d3d12_resource_pool_texture.cpp
```diff
@@ -542,6 +542,235 @@ namespace wr
 		return texture;
 	}
 
+	d3d12::TextureResource * D3D12TexturePool::LoadPNGFromMemory(char * data, size_t size, bool srgb, bool generate_mips)
```
LaisoEmilio: I didn't see this in the first review, but the load functions have been modified recently, there's some pieces of code that you are missing from this one. An example is the code that adjustes the format that you can see in the LoadPNG function, also the set name call on the texture resouce. If you don't have time to do this, write it in a comment to this one and I'll do it as soon as I have time

PR #171, opened 2019-02-21
Feature soft shadows

MeineZ:
Soft-shadows that take 4 samples per light (per reflection). 
The number of samples can be changed in shadow_ray.hlsl.

**!!! BREAKS FALLBACK ON NON-RTX MACHINES SINCE SHADOWS ARE FUNCTIONAL !!!**

**Suggested improvements (displayed in code with comments too)**
- **Lighting.hlsl, line 90:** Maybe change hard-coded 100000 to be dynamic according to the scene size? Not sure if length of rays improve performance though..
- **Lighting.hlsl, line 104:** Perhaps change randomness to not be 'purely' random, but algorithm-random?
- **Lighting.hlsl, line 107:** Hard-coded 0.05 is to correct the offset a ray gets. Should be determined by the area that the light is emitting from.


resources/shaders/shadow_ray.hlsl
```diff
@@ -1,5 +1,7 @@
 #include "rt_global.hlsl"
 
+#define MAX_SHADOW_SAMPLES 4
```
VZout: Move this to rt_global please.

resources/shaders/lighting.hlsl
```diff
@@ -85,20 +85,41 @@ float3 shade_light(float3 pos, float3 V, float3 albedo, float3 normal, float met
...
 	float3 lighting = BRDF(L, V, normal, metallic, roughness, albedo, radiance);
 
-	bool shadow = TraceShadowRay(1, pos + (normal * EPSILON), L, t_max, depth + 1);
-	lighting *= !shadow;
+	float3 wpos = pos + (normal * EPSILON);
+
+	// Offset shadow ray direction to get soft-shadows
+	float shadow_factor = 0.0;
+	[unroll]
+	for (uint i = 0; i < MAX_SHADOW_SAMPLES; ++i)
```
VZout: Can you add a define to disable soft shadows?

demo/scene_viknell.hpp
```diff
@@ -19,7 +21,8 @@ namespace viknell_scene
 	void CreateScene(wr::SceneGraph* scene_graph, wr::Window* window)
 	{
 		camera = scene_graph->CreateChild<DebugCamera>(nullptr, 90.f, (float)window->GetWidth() / (float)window->GetHeight());
-		camera->SetPosition({ 0, 0, 2 });
+		camera->SetPosition({ 0.35, -0.1, 1.3 });
```
VZout: Please revert your changes to the demo scene. Its better to keep it consistent to if we need to go back in the commit history comparing is easy.

PR #174, opened 2019-02-22
Changed imgui tools to be able to manipulate models

161563:
[CHANGE] Added model selector with position gizmo
[CHANGE] Modified inspector widget to also display model data


VZout:
Also. Instead of having a 2e window for the model editor edit the scene graph editor to contain all nodes. And rename it to scene graph editor. Every node extends the same base with transform. so if we add a new type of node it automatically works with the inspector.

PR #181, opened 2019-03-01
Support for copying data to non 256-byte aligned buffer sizes

tntmeijs:
We've always tested the plug-in using a resolution of 1280*720. This is a perfect multiple of 256 when calculating the row pitch of a buffer by doing width*bytes per pixel.

This pull request adds support for non perfect 256-byte row pitches. The code now adds padding whenever needed.

### DIFF
--- a/src/util/math.hpp
+++ b/src/util/math.hpp
@@ -18,20 +18,4 @@ namespace util
 
 		return upper3x3;
 	}
-
-	// https://stackoverflow.com/a/3407254
-	//! Round the input number to the nearest multiple of the specified number
-	/*! \param input Number to round. 
-	 *  \param multiple Multiple to round the input to. */
-	inline std::uint32_t RoundUpToNearestMultiple(std::uint32_t input, std::uint32_t multiple)
-	{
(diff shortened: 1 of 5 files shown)


src/util/math.hpp
```diff
@@ -19,4 +19,19 @@ namespace util
 		return upper3x3;
 	}
 
+	// https://stackoverflow.com/a/3407254
+	//! Round the input number to the nearest multiple of the specified number
+	/*! \param input Number to round. 
+	 *  \param multiple Multiple to round the input to. */
+	inline std::uint32_t RoundUpToNearestMultiple(std::uint32_t input, std::uint32_t multiple)
```
VZout: We have a size align function some where I believe. d3d12_defines.hpp. its a shitty location I know.

florianschut:
Didn't test the code but it looks solid to me! 👍

src/util/math.hpp
tntmeijs: I'll change to code to make use of that function instead. Thanks!

PR #187, opened 2019-03-12
Feature dynamic vertex buffers

161563:
Model pools can now be resized and defragmented, and models can have their data changed after creation. Model pools now also automatically defragment and resize when not enough memory is available.
Added the ability to rebuild altered model acceleration structures if necessary, as well as rebuilding the top-level structure when additional models/bottom-level structures are added.


src/model_pool.hpp
```diff
@@ -185,6 +246,25 @@ namespace wr
 
 		// TODO: Create default material
 
+		size_t total_vertex_size = 0;
```
VZout: Duplicate code here

src/model_pool.hpp
```diff
@@ -397,6 +479,25 @@ namespace wr
 			material_handles.push_back(new_handle);
 		}
 
+		size_t total_vertex_size = 0;
```
VZout: duplicate code here.

src/render_tasks/d3d12_build_acceleration_structures.hpp
```diff
@@ -298,6 +434,8 @@ namespace wr
 
 			data.out_materials_require_update = false;
 
+			data.current_frame_index = n_render_system.GetFrameIdx();
```
VZout: Probably don't want to store the current frame index in the data. Doesn't make sense to me.

src/d3d12/d3d12_acceleration_structure..cpp
```diff
@@ -123,6 +123,7 @@ namespace wr::d3d12
 		std::vector<desc::GeometryDesc> geometry)
 	{
 		AccelerationStructure blas;
+		ZeroMemory(&blas, sizeof(blas));
```
VZout: We are not programming `C` Constructors should deal with potential errors here.

src/d3d12/d3d12_acceleration_structure..cpp
```diff
@@ -223,6 +224,7 @@ namespace wr::d3d12
 		std::vector<std::tuple<d3d12::AccelerationStructure, unsigned int, DirectX::XMMATRIX>> blas_list)
 	{
 		AccelerationStructure tlas;
+		ZeroMemory(&tlas, sizeof(tlas));
```
VZout: Still not programming C :)

src/d3d12/d3d12_defines.hpp
```diff
@@ -25,4 +25,10 @@ template<typename T, typename A>
 constexpr inline T SizeAlign(T size, A alignment)
 {
 	return (size + (alignment - 1U)) & ~(alignment - 1U);
-}
\ No newline at end of file
+}
+
+template<typename T, typename A>
+constexpr inline T SizeAlignProper(T size, A alignment)
```
VZout: Poor naming. Use comments and rename one or both so they actually make sense.

src/d3d12/d3d12_model_pool.hpp
```diff
@@ -32,6 +32,51 @@ namespace wr
...
+
+		enum CommandType
+		{
+			STAGE,
+			COPY,
+			READ,
+			TRANSITION,
+		};
+
+		struct Command
```
VZout: Unused?

src/d3d12/d3d12_staging_buffer.cpp
```diff
@@ -12,13 +12,13 @@ namespace wr::d3d12
 		buffer->m_target_resource_state = resource_state;
 		buffer->m_size = size;
 		buffer->m_stride_in_bytes = stride;
-		buffer->m_is_staged = false;
+		buffer->m_is_staged = true;
```
VZout: Sure about this change?

src/model_pool.hpp
```diff
@@ -116,6 +154,27 @@ namespace wr
 
 		auto model = new Model();
 
+		std::size_t total_vertex_size = 0;
```
VZout: Duplicate code here.

src/model_pool.hpp
```diff
@@ -411,8 +512,32 @@ namespace wr
...
+	{
+		UpdateMeshData(mesh,
+			vertices.data(),
+			vertices.size(),
+			sizeof(TV),
+			indices.data(),
+			indices.size(),
+			sizeof(TI));
+
+		for (int i = 0; i < m_loaded_models.size(); i++)
```
VZout: prefer range based loops for readability.

src/d3d12/d3d12_model_pool.hpp
161563: Functions as a common ancestor for all different commands, so they can all fit into a single queue.

src/d3d12/d3d12_staging_buffer.cpp
161563: Yeah, to my knowledge this isn't used anywhere, and was just something so a buffer wouldn't get staged twice. But this isn't necessary anymore after staging buffer rewrites.

src/render_tasks/d3d12_build_acceleration_structures.hpp
161563: I need the data in the other helper functions as well, and they can't access the render_system. This seemed better than modifying those functions to accept the render system as a variable and calling the GetFrameIdx function each time.

161563:
Applied feedback.
Fixed an issue that arose when editing a mesh, where the pointer to the material handle would become invalid. Fixed this by making the acceleration structure stop using pointer for material handles, like the rest of the system.
Added methods for testing editing, resizing and defragmenting. Hotkeys are F3 for defragmenting, F4 for shrink to fit, and F5 for editing a mesh, editing one mesh of the xbot model into a part of the material ball model.

PR #195, opened 2019-03-19
Image-based lighting

LaisoEmilio:
Adding image based lighting to the engine.
Also changed a massive amount of code non related to IBL, like adding names to render targets for the render tasks and stuff like that. My bad, I should've separated the changes into different branches.


demo/engine_interface.hpp
```diff
@@ -131,6 +131,23 @@ namespace engine
 			ImGui::End();
 		}
 
+		{
+			ImGui::Begin("IBL Testing Ground");
```
VZout: This should go

demo/resources.hpp
```diff
@@ -117,7 +119,11 @@ namespace resources
 		mirror_internal->SetAlbedo(white);
 		mirror_internal->SetNormal(flat_normal);
 		mirror_internal->SetRoughness(black);
+		mirror_internal->UseRoughnessTexture(false);
+		mirror_internal->SetConstantRoughness(0.0f);
```
VZout: This causes the difference between deferred/hybrid & fullraytracing. That constant value doesn't work properly.

PR #206, opened 2019-03-25
[FIX] Fixed soft shadows not working on fallback (hopefully)

161563:
Removed loop unroll, which should fix the crashes we're getting with soft shadows on fallback.

### DIFF
--- a/resources/shaders/rt_global.hlsl
+++ b/resources/shaders/rt_global.hlsl
@@ -4,11 +4,11 @@
 //#define DEPTH_OF_FIELD
 #define EPSILON 0.01
 #define SOFT_SHADOWS
-#define MAX_SHADOW_SAMPLES 4
+#define MAX_SHADOW_SAMPLES 3
 
 #ifdef FALLBACK
 	#undef MAX_RECURSION
-	#define MAX_RECURSION 1
+	#define MAX_RECURSION 3
 
 	#define NO_SHADOWS
 #endif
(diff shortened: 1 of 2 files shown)


florianschut:
Still breaks fallback

resources/shaders/rt_global.hlsl
```diff
@@ -3,11 +3,12 @@
...
 //#define DEPTH_OF_FIELD
 #define EPSILON 0.01
-#define MAX_SHADOW_SAMPLES 4
+#define SOFT_SHADOWS
+#define MAX_SHADOW_SAMPLES 3
 
 #ifdef FALLBACK
 	#undef MAX_RECURSION
-	#define MAX_RECURSION 1
+	#define MAX_RECURSION 3
```
Nielsbishere: Does fallback run at reasonable speeds with this? Or only with 1 recursion?

resources/shaders/shadow_ray.hlsl
```diff
@@ -10,7 +10,7 @@ struct Attributes { };
 
 bool TraceShadowRay(uint idx, float3 origin, float3 direction, float far, unsigned int depth)
 {
-#ifndef NO_SHADOWS
+//#ifndef NO_SHADOWS
```
Nielsbishere: If you can already disable/enable shadows through other methods, this has to be removed. Otherwise this should be left here and not commented.

resources/shaders/shadow_ray.hlsl
```diff
@@ -60,7 +60,6 @@ float GetShadowFactor(float3 wpos, float3 light_dir, float t_max, uint depth, in
 	float shadow_factor = 0.0f;
 
 #ifdef SOFT_SHADOWS
-	[unroll(MAX_SHADOW_SAMPLES)]
```
Nielsbishere: Why remove this unroll? Doesn't that increase performance?

PR #208, opened 2019-03-25
Fixed issues that came up when upgrading Wisp to latest master

tntmeijs:
The demo does not make use of these two render tasks, so nobody actually noticed it.
I just updated the code to compile against the latest Wisp master.


tntmeijs:
The name has been added to the render tasks again.

PR #217, opened 2019-03-29
Feature soft shadows reflections fallback fix

161563:
Fixes shadows and reflections crashing the engine on the fallback layer.

Note: Having recursive reflections and shadows on at the same time destroys fps on fallback. Using a recursion level higher than 2 seems to cause some weird artifacts (not sure if this is because of the fix, fallback or just what it's supposed to do).


resources/shaders/shadow_ray.hlsl
```diff
@@ -10,7 +10,7 @@ struct Attributes { };
 
 bool TraceShadowRay(uint idx, float3 origin, float3 direction, float far, unsigned int depth)
 {
-#ifndef NO_SHADOWS
+//#ifndef NO_SHADOWS
```
Nielsbishere: Conflicts with this and your previous branch; if shadows can be disabled runtime already, this should be completely removed instead of commented. If they can only be disabled compile time, please leave this in (for benchmarking etc.).

resources/shaders/shadow_ray.hlsl
```diff
@@ -60,7 +60,6 @@ float GetShadowFactor(float3 wpos, float3 light_dir, float t_max, uint depth, in
 	float shadow_factor = 0.0f;
 
 #ifdef SOFT_SHADOWS
-	[unroll(MAX_SHADOW_SAMPLES)]
```
Nielsbishere: Conflicts with this and your previous branch; why is this unroll removed, does it not increase performance?

161563: It crashes fallback.

VZout:
Closing this PR since it doesn't seem to fix it anymore.

PR #232, opened 2019-04-02
Feature fix dds miplevels

LaisoEmilio:
Reworked the texture pool class a bit and fixed the DDS mipmaps issue


src/resource_pool_texture.cpp
```diff
@@ -9,94 +9,20 @@ namespace wr
...
 		});
 
 		TextureType type;
-		if (texture_extension.size() == 0 || height > 1)
-		{
-			type = TextureType::RAW;
-		}
-		else if (texture_extension.compare("png") == 0)
+
+		if (texture_extension.compare("png") == 0
```
VZout: Why use std::compare here? its slower here than a normal string comp and is less readable.

src/resource_pool_texture.hpp
```diff
@@ -32,9 +32,10 @@ namespace wr
...
-		[[nodiscard]] TextureHandle LoadFromMemory(char* data, int width, int height, std::string texture_extension, bool srgb, bool generate_mips);
+		[[nodiscard]] virtual TextureHandle LoadFromFile(std::string_view path, bool srgb, bool generate_mips) = 0;
+		[[nodiscard]] virtual TextureHandle LoadFromCompressedMemory(char* data, size_t width, size_t height, std::string& texture_extension, bool srgb, bool generate_mips);
+		[[nodiscard]] virtual TextureHandle LoadFromCompressedMemory(char* data, size_t width, size_t height, TextureType type, bool srgb, bool generate_mips) = 0;
+		[[nodiscard]] virtual TextureHandle LoadFromRawMemory(char* data, size_t width, size_t height, bool srgb, bool generate_mips) = 0;
 		[[nodiscard]] virtual TextureHandle CreateCubemap(std::string_view name, uint32_t width, uint32_t height, uint32_t mip_levels, Format format, bool allow_render_dest) = 0;
 		[[nodiscard]] virtual TextureHandle CreateTexture(std::string_view name, uint32_t width, uint32_t height, uint32_t mip_levels, Format format, bool allow_render_dest) = 0;
 		virtual void Unload(uint64_t texture_id) = 0;
```
VZout: Why does this use the id and not the entire handle :/

src/resource_pool_texture.hpp
```diff
@@ -32,9 +32,10 @@ namespace wr
 		TexturePool(TexturePool&&) = delete;
 		TexturePool& operator=(TexturePool&&) = delete;
 
-		[[nodiscard]] TextureHandle Load(std::string_view path, bool srgb, bool generate_mips);
-		[[nodiscard]] TextureHandle LoadFromMemory(char* data, int width, int height, TextureType type, bool srgb, bool generate_mips);
-		[[nodiscard]] TextureHandle LoadFromMemory(char* data, int width, int height, std::string texture_extension, bool srgb, bool generate_mips);
+		[[nodiscard]] virtual TextureHandle LoadFromFile(std::string_view path, bool srgb, bool generate_mips) = 0;
+		[[nodiscard]] virtual TextureHandle LoadFromCompressedMemory(char* data, size_t width, size_t height, std::string& texture_extension, bool srgb, bool generate_mips);
```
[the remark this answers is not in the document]
LaisoEmilio: Changed it

src/resource_pool_texture.cpp
LaisoEmilio: 👍

src/resource_pool_texture.hpp
LaisoEmilio: Oversight! changed it

LaisoEmilio: Changed this, and much more

src/resource_pool_texture.cpp
```diff
@@ -7,28 +7,33 @@ namespace wr
 
 	TexturePool::TexturePool()
 	{
+#ifdef _DEBUG
+		static uint16_t pool_count = 0u;
```
VZout: The render system should be in charge of creating pools. You could make the constructor of pool private and friend it with render system. so you can ONLY create it using the RS. that way you can count the number of pools from that list.,

src/resource_pool_texture.hpp
```diff
@@ -52,7 +52,7 @@ namespace wr
 		TextureHandle GetDefaultMetalic();
 		TextureHandle GetDefaultAO();
 
-		virtual Texture* GetTexture(uint64_t texture_id) = 0;
+		virtual Texture* GetTextureResource(TextureHandle handle) = 0;
```
VZout: you could template this so it casts internally. instead of having to cast it every time you call this function.

VZout: Also this function technically shouldn't have to be pure virtual. it could be implemented here if written well.

src/render_tasks/d3d12_equirect_to_cubemap.hpp
```diff
@@ -112,8 +112,8 @@ namespace wr
 			//Get allocator from pool
 			auto allocator = static_cast<D3D12TexturePool*>(src_texture.m_pool)->GetMipmappingAllocator();
 
-			d3d12::TextureResource* cubemap_text = static_cast<d3d12::TextureResource*>(src_texture.m_pool->GetTexture(src_texture.m_id));
-			d3d12::TextureResource* envmap_text = static_cast<d3d12::TextureResource*>(dst_texture.m_pool->GetTexture(dst_texture.m_id));
+			d3d12::TextureResource* cubemap_text = static_cast<d3d12::TextureResource*>(src_texture.m_pool->GetTextureResource(src_texture));
```
DBouma: This seems to happen more with your naming, the "_text" suffix is slightly confusing for me. This is a minor thing though

PR #233, opened 2019-04-02
Feature scene graph editor

VZout:
- New scene graph editor that shows all nodes in a list.
- Better naming for nodes.
- More abstraction and extend-ability.
- Nodes are now actually properly removed from the scene graph when deleted.

Still needs a bit of refactoring but it is already much better than the previous implementation.

This PR also contains:

- Better chromatic aberration.
- Barrel distortion
- FXAA

### DIFF
--- a/imgui.ini
+++ b/imgui.ini
@@ -21,14 +21,14 @@ Collapsed=0
 DockId=0x00000008,0
 
 [Window][Camera Settings]
-Pos=0,47
-Size=11,6
+Pos=0,41
+Size=11,12
 Collapsed=0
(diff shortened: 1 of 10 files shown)


imgui.ini
```diff
@@ -21,14 +21,14 @@ Collapsed=0
 DockId=0x00000008,0
```
tntmeijs: Are you sure you want to submit this file?

VZout: Yep! I removed some windows and added a new one. So I need to update the layout.

PR #234, opened 2019-04-04
Reduced equirect size from 4k to 1k

Nielsbishere:
Since it requires 4 extra GiB, which dev's (and most regular devices) can't afford, since we're already at 5 GiB.

### DIFF
--- a/src/render_tasks/d3d12_equirect_to_cubemap.hpp
+++ b/src/render_tasks/d3d12_equirect_to_cubemap.hpp
@@ -180,9 +180,9 @@ namespace wr
 
 			data.in_equirect = skybox_node->m_hdr;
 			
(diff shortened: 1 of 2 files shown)


src/render_tasks/d3d12_equirect_to_cubemap.hpp
```diff
@@ -180,7 +180,7 @@ namespace wr
 
 			data.in_equirect = skybox_node->m_hdr;
 			
-			skybox_node->m_skybox = skybox_node->m_hdr.m_pool->CreateCubemap("Skybox", 3840, 3840, 0, wr::Format::R32G32B32A32_FLOAT, true);
+			skybox_node->m_skybox = skybox_node->m_hdr.m_pool->CreateCubemap("Skybox", 1024, 1024, 0, wr::Format::R32G32B32A32_FLOAT, true);
```
VZout: Make this a d3d12_settings variable plz :)

Nielsbishere: Done

Issue #230, opened 2019-04-02
Material has duplicated flags


Nielsbishere:
**Is your request related to a problem? Please describe.**
Material has flags such as "has_roughness_texture" and "has_roughness_constant", these flags aren't required, since if there's no texture it will use the constant.

**Describe the solution you'd like**
Remove has_x_constant.

**Describe alternatives you've considered**
**Additional context**
N.A.

Nielsbishere:
Material doesn't even need these flags (only on the GPU). You could just check if it has a textureptr

Issue #222, opened 2019-04-01
Duplicated code


Nielsbishere:
**Is your design request related to a problem? Please describe.**
A lot of code that does almost exactly the same thing could be prevented by using templates or by smarter design decisions. Examples of these cases:
```
material_pool
model_pool
```

**Describe the solution you'd like**
When using something like setting textures, use an enum to index into a `TextureHandle[]` instead of duplicating the same function and rewriting the same thing a few times. 
Template functions could check for a field to be present in a class, therefore allowing you to do something like:
```cpp
if constexpr(HasField(MyVertexFormat, pos))
  ; //Copy into pos, since it exists
```
_Note that you'd have to define HasField yourself_.

**Describe alternatives you've considered**
N.A.

**Additional context**
N.A.

Nielsbishere:
I have a PoC for HasField:
```cpp
#define HAS_FIELD(funcName, name, ...)																	\
template<typename, typename, typename = int>															\
struct T##funcName {																					\
	static constexpr bool value = false;																\
};																										\
																										\
template<typename T, typename Ret>																		\
struct T##funcName<T, Ret, 																				\
std::enable_if_t<std::is_same_v<Ret, decltype(T::name)>, int>											\
> {																										\
	static constexpr bool value = true;																	\
};																										\
																										\
template<typename T>																					\
struct funcName {																						\
	static constexpr bool value = T##funcName<T, __VA_ARGS__, int>::value;								\
};
```
This can be used like the following:
```cpp

struct MyType { };
struct MyType0 { float field;  };

HAS_FIELD(HasField, field, float);

static constexpr bool bools[2] = { HasField<MyType>::value, HasField<MyType0>::value };
```
Meaning that you can type;
```cpp
namespace wr::magic {
HAS_FIELD(HasPosition, pos, XMFLOAT3);
}

//model loader
if constexpr(wr::magic::HasPosition<T>::value) 
 ; //Load into position buffer
```

VZout:
I'm not sure what code you exactly mean here. Can you give some line numbers?

Nielsbishere:
For example; in model pool, there's the sample code copy pasted 5x one for every vertex type. We could have checks for for example position and then only call the set function if it has it. Makes it easier to modify as well.

Issue #212, opened 2019-03-28
Material Unload


MeineZ:
Materials can't be individually removed from the material pool.

**Describe the solution you'd like**
Being able to remove materials individually.

Issue #209, opened 2019-03-26
Hybrid & Path tracing doesn't work with material constants if you don't set a texture as well.


VZout:
If you set a material constant and don't set a texture the renderer will crash on initialization.

PR #238, opened 2019-04-04
Feature refactor materials

Nielsbishere:
Materials are now refactored and don't store duplicated data / don't have duplicated functions. This fixes textureless materials for RT.


Nielsbishere:
https://github.com/TeamWisp/Procedural-Ray-Tracing/issues/230
https://github.com/TeamWisp/Procedural-Ray-Tracing/issues/222
https://github.com/TeamWisp/Procedural-Ray-Tracing/issues/209

Nielsbishere:
https://github.com/TeamWisp/Procedural-Ray-Tracing/issues/212

src/material_pool.hpp
```diff
@@ -37,130 +52,83 @@ namespace wr
...
+		template<uint16_t s>
+		void GetConstant(MaterialConstantType type, float(&val)[s]);
+		
+		template<uint16_t s>
+		void SetConstant(MaterialConstantType type, float(&val)[s]);
 
-		TextureHandle GetRoughness() { return m_rougness; }
-		void SetRoughness(TextureHandle roughness);
+		void SetConstant(MaterialConstantType type, float val);
+		void GetConstant(MaterialConstantType type, float &val);
```
VZout: Why is this a reference and not a return value. Are we writing C now again?

src/material_pool.hpp
```diff
@@ -37,130 +52,83 @@ namespace wr
...
-
-		TextureHandle GetMetallic() { return m_metallic; }
-		void SetMetallic(TextureHandle metallic);
-
-		void UseMetallicTexture(bool use_metallic);
+		TexturePool* const GetTexturePool() { return m_texture_pool; }
 
-		TextureHandle GetAmbientOcclusion() { return m_ao; }
-		void SetAmbientOcclusion(TextureHandle ao);
+		ConstantBufferHandle* GetConstantBufferHandle() { return m_constant_buffer_handle; };
```
VZout: Should have moved the impl to the cpp file and marked it const.

src/material_pool.hpp
```diff
@@ -37,130 +52,83 @@ namespace wr
...
-			{
-				DirectX::XMVectorGetX(m_material_data.m_metallic_roughness),
-				DirectX::XMVectorGetY(m_material_data.m_metallic_roughness),
-				DirectX::XMVectorGetZ(m_material_data.m_metallic_roughness)
-			};
+			uint32_t m_value;
 		};
 
-		void SetConstantMetallic(DirectX::XMFLOAT3 metallic);
+		union MaterialData
```
VZout: Does this mean we can't have an albedo texture and constants for the rest?

Nielsbishere: Nope, it should still work exactly the same, just refactored and 'setConstant' and 'setTexture' are now used. You can just have an albedo texture and constants for everything else, I've tested it with pure constants and a mix and it worked fine.

src/material_pool.hpp
Nielsbishere: It's just a getter, I agree with the const, but idk if having an implementation in the CPP for just a return is required.

src/material_pool.hpp
Nielsbishere:
```cpp
		template<uint16_t s>
		void GetConstant(MaterialConstantType type, float(&val)[s]);
```
This uses a float array to return the data into, since you can't return that. The GetConstant with a float is just an overload to that (consistency).

demo/scene_emibl.hpp
```diff
@@ -106,103 +106,103 @@ namespace emibl_scene
 
 				wr::Material* rusty_metal_internal = material_pool->GetMaterial(rusty_metal_material);
 
-				rusty_metal_internal->SetAlbedo(metal_splotchy_albedo);
```
VZout: asd

PR #241, opened 2019-04-05
Feature refactor skybox

Nielsbishere:
This refactors the render tasks to use the TextureCube converted from the equirect map instead of converting direction to 2D coords on a 2D texture. It also cleans up the intermediate equirect texture, cleaning up ~0.2 GiB.


src/render_tasks/d3d12_path_tracer.hpp
```diff
@@ -329,30 +329,21 @@ namespace wr
...
 				// Get skybox
-				if (scene_graph.m_skybox.has_value())
-				{
-					auto skybox_t = static_cast<d3d12::TextureResource*>(scene_graph.m_skybox.value().m_pool->GetTextureResource(scene_graph.m_skybox.value()));
-					d3d12::SetRTShaderSRV(cmd_list, 0, COMPILATION_EVAL(rs_layout::GetHeapLoc(params::path_tracing, params::PathTracingE::SKYBOX)), skybox_t);
-				}
-
-				// Get Pre-filtered environment
-				if (scene_graph.m_skybox.has_value())
+				if (SkyboxNode *skybox = scene_graph.GetCurrentSkybox().operator->())
```
VZout: `.operator->()` I don't think this is necessary

Nielsbishere: Replaced with .get() instead, intellisense didn't show it as an option.

VZout:
conflicts :o

PR #253, opened 2019-04-13
Feature splines

VZout:
* Added a spline library to demo
* Viknell has a single camera spline
* node.hpp now requires a typeid to be initialized for "reflection"
* node.hpp can now use quaternions directly.


VZout:
Forcing merge since it contains some very important changes for materials.

PR #272, opened 2019-05-07
Direct GI

VZout:
Added direct GI from before the vacation.

### DIFF
--- a/resources/shaders/accumulation.hlsl
+++ b/resources/shaders/accumulation.hlsl
@@ -17,11 +17,7 @@ void main(uint3 DTid : SV_DispatchThreadID)
 	float2 resolution;
 	input.GetDimensions(resolution.x, resolution.y);
 	
-	float2 uv = float2(DTid.xy) / resolution;
-	float gamma = 2.2;
-	float exposure = 1;
-
-	float3 color = input[DTid.xy].rgb / (frame_idx);
+	float3 color = input[DTid.xy].rgb / (frame_idx + 1);
 
 	output[DTid.xy] = float4(color, 1);
 }
--- a/resources/shaders/path_tracer.hlsl
+++ b/resources/shaders/path_tracer.hlsl
@@ -3,6 +3,7 @@
 #include "pbr_util.hlsl"
 #include "material_util.hlsl"
 #include "lighting.hlsl"
+#include "gi_util.hlsl"
 
 struct Vertex
 {
@@ -47,14 +48,6 @@ SamplerState s0 : register(s0);
 
 typedef BuiltInTriangleIntersectionAttributes MyAttributes;
 
-struct HitInfo
-{
-	float3 color;
-	unsigned int seed;
-	float3 origin;
-	unsigned int depth;
-};
-
 cbuffer CameraProperties : register(b0)
 {
 	float4x4 inv_view;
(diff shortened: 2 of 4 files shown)


resources/shaders/path_tracer.hlsl
```diff
@@ -139,6 +139,66 @@ float probabilityToSampleDiffuse(float3 difColor, float3 specColor)
...
+	float3 F = F_Schlick(LdotH, metal, albedo);
+	//float3 F = F_ShlickSimple(metal, LdotH);
+
+	float3 kS = F_SchlickRoughness(NdotV, metal, albedo, roughness);
+	float3 kD = (1.f - kS) * (1.0 - metal);
+	float3 spec = (D * F * G) / (4.0 * NdotV * NdotL + 0.001f);
+
+	//prefiltered_color * (kS * sampled_brdf.x + sampled_brdf.y);
+	return shadow_mult * (light_intensity * (NdotL * spec + NdotL * albedo / M_PI));
+	return shadow_mult * (kD * albedo / PI + spec) * light_intensity * NdotL;
```
Nielsbishere: Why is there two returns here?

resources/shaders/path_tracer.hlsl
```diff
@@ -96,7 +96,7 @@ float4 TraceColorRay(float3 origin, float3 direction, unsigned int depth, unsign
 	RayDesc ray;
 	ray.Origin = origin;
 	ray.Direction = direction;
-	ray.TMin = EPSILON;
+	ray.TMin = 0;
```
Nielsbishere: Won't this cause self-intersection?

resources/shaders/accumulation.hlsl
```diff
@@ -17,11 +17,7 @@ void main(uint3 DTid : SV_DispatchThreadID)
 	float2 resolution;
 	input.GetDimensions(resolution.x, resolution.y);
 	
-	float2 uv = float2(DTid.xy) / resolution;
-	float gamma = 2.2;
-	float exposure = 1;
-
-	float3 color = input[DTid.xy].rgb / (frame_idx);
+	float3 color = input[DTid.xy].rgb / (frame_idx + 1);
```
Nielsbishere: Why do you divide by frame_idx? Means the color would be divided by 1, then 2, then 3?

resources/shaders/path_tracer.hlsl
```diff
@@ -139,6 +139,66 @@ float probabilityToSampleDiffuse(float3 difColor, float3 specColor)
 	return lumDiffuse / (lumDiffuse + lumSpecular);
 }
 
+float3 ggxDirect(float3 hit_pos, float3 fN, float3 N, float3 V, float3 albedo, float metal, float roughness, unsigned int seed, unsigned int depth)
+{
+	// #################### GGX #####################
+	uint light_count = lights[0].tid >> 2;
+	int light_to_sample = min(int(nextRand(seed) * light_count), light_count - 1);
+	Light light = lights[light_to_sample];
```
Nielsbishere: What if there's no lights available?

resources/shaders/path_tracer.hlsl
Nielsbishere: Nvm, see you added it to the origin

resources/shaders/accumulation.hlsl
VZout: Yes

resources/shaders/path_tracer.hlsl
VZout: Good point. Forgot about that case.

resources/shaders/path_tracer.hlsl
VZout: removed

resources/shaders/accumulation.hlsl
Nielsbishere: Hmm, I guess the var name is correct, but frame_idx is also referenced to as the current render index

PR #273, opened 2019-05-07
Framegraph constructor moved to the source file

tntmeijs:
The linker throws the "Unresolved External Symbol" error when the constructor is in the header.


tntmeijs:
Done, everything is now ready to be merged.

VZout:
I dislike this change. The static varaibles should just be inlined.

PR #274, opened 2019-05-07
Save Render targets to disc.

VZout:
* Ability to request a save of a render target.
* Cleaned the frame graph slightly
* In-lined the statics of the frame graph. Hopefully solves #273


src/d3d12/d3d12_renderer.cpp
```diff
@@ -499,6 +507,83 @@ namespace wr
...
+
+		auto cmd_list = d3d12::CreateCommandList(m_device, 1, CmdListType::CMD_LIST_DIRECT);
+		SetName(cmd_list, L"Screenshot Command List");
+
+		auto fence = d3d12::CreateFence(m_device);
+
+		d3d12::Begin(cmd_list, 0);
+
+		// Readback
+		auto format_size = d3d12::SizeOfFormat(format);
```
tntmeijs: Don't we have a readback buffer implementation already?

src/d3d12/d3d12_textures.cpp
```diff
@@ -725,4 +725,126 @@ namespace wr::d3d12
 		return out_format;
 	}
 
+	std::size_t SizeOfFormat(Format format)
```
tntmeijs: I remember adding this to some utility file a few months ago. Is this duplicate code, or was the old function removed?

PR #283, opened 2019-05-09
Refactored render tasks to better use the descriptor heap.

LaisoEmilio:
The previous version of deferred composition made no sense, it was hard to understand what was going on. This version adds some lines of code, but at least it makes it clearer.


src/render_tasks/d3d12_deferred_composition.cpp
```diff
@@ -30,21 +30,21 @@ namespace wr
 
 			d3d12::BindComputeConstantBuffer(cmd_list, camera_cb, 0, frame_idx);
 
-			constexpr unsigned int albedo = rs_layout::GetHeapLoc(params::deferred_composition, params::DeferredCompositionE::GBUFFER_ALBEDO_ROUGHNESS);
-			d3d12::DescHeapCPUHandle albedo_handle = data.out_rtv_srv_allocation.GetDescriptorHandle(albedo);
-			d3d12::SetShaderSRV(cmd_list, 1, albedo, albedo_handle);
+			constexpr unsigned int albedo_loc = rs_layout::GetHeapLoc(params::deferred_composition, params::DeferredCompositionE::GBUFFER_ALBEDO_ROUGHNESS);
```
DBouma: Consider writing albedo_heap_loc instead of albedo_loc, it was not clear what this "_loc" was to me.

DBouma: Same goes for the other values

PR #285, opened 2019-05-09
Feature bloom new

DBouma:
* added bloom 

sigma is hardcoded in shader, its a basic version that will be improved over the next few days.


resources/shaders/bloom_composition.hlsl
```diff
@@ -0,0 +1,35 @@
...
+	float3 finalcolor = float3(0, 0, 0);
+	float bloom_intensity = 1.f;
+
+	if (enable_bloom > 0)
+	{
+		finalcolor += source_bloom.SampleLevel(linear_sampler, uv, 0).rgb * bloom_intensity;
+	}
+	finalcolor += source_main.SampleLevel(point_sampler, uv, 0).rgb;
+
+	output[int2(dispatch_thread_id.xy)] = float4(finalcolor, 1.0f);
```
VZout: Bloom shouldn't require a composition. Just task just should take a input and use that input also as output.

src/render_tasks/d3d12_bloom_composition.hpp
```diff
@@ -0,0 +1,181 @@
...
+{
+
+	struct BloomSettings
+	{
+		struct Runtime
+		{
+			float m_sigma = 2.0f;
+			float m_bloom_intensity = 1.0f;
+			float m_luminance_threshold = 1.0f;
+			bool m_enable_bloom = true;
```
VZout: Disabling bloom should be the same as not using the bloom render task.

PR #289, opened 2019-05-10
Cleanup and now using uints for the rt materials and offsets

### LINKED ISSUE #288: Use uints when needed
**Is your design request related to a problem? Please describe.**
Floats are used instead of integers for texture handles and things like offsets. This is not only bad practice but also results in problems when we reach >8M vertices or indices. 

**Describe the solution you'd like**
Whenever using integer data, use a uint; don't use a float just cause it's easy.

**Describe alternatives you've considered**
**Additional context**
N.A.

VZout:
Fixes #288

### DIFF
--- a/src/d3d12/d3d12_command_list.cpp
+++ b/src/d3d12/d3d12_command_list.cpp
@@ -12,7 +12,7 @@ namespace wr::d3d12
 	{
 		auto* cmd_list = new CommandList();
 		const auto n_device = device->m_native;
-		const auto n_cmd_list = cmd_list->m_native;
+		auto& n_cmd_list = cmd_list->m_native;
 
 		cmd_list->m_allocators.resize(num_allocators);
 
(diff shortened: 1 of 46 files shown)


src/d3d12/d3d12_command_list.cpp
```diff
@@ -115,7 +115,7 @@ namespace wr::d3d12
 		CD3DX12_CPU_DESCRIPTOR_HANDLE dsv_handle;
 		if (render_target->m_create_info.m_create_dsv_buffer) dsv_handle = render_target->m_depth_stencil_resource_heap->GetCPUDescriptorHandleForHeapStart();
 
-		cmd_list->m_native->OMSetRenderTargets(handles.size(), handles.data(), false, render_target->m_create_info.m_create_dsv_buffer ? &dsv_handle : nullptr);
+		cmd_list->m_native->OMSetRenderTargets(static_cast<unsigned int>(handles.size()), handles.data(), false, render_target->m_create_info.m_create_dsv_buffer ? &dsv_handle : nullptr);
```
tntmeijs: You're using UINT and unsigned int, should we stick to one style?

src/d3d12/d3d12_command_list.cpp
```diff
@@ -174,18 +174,18 @@ namespace wr::d3d12
 
 	void BindDescriptorHeap(CommandList* cmd_list, DescriptorHeap* heap, DescriptorHeapType type, unsigned int frame_idx, bool fallback)
 	{
-		size_t heap_idx = frame_idx % heap->m_create_info.m_versions;
+		std::uint32_t heap_idx = frame_idx % heap->m_create_info.m_versions;
```
tntmeijs: Yet another way of writing a uint

VZout:
replaced all uints with stduint "D

PR #301, opened 2019-05-16
Cleanup and performance improvements.

VZout:
* We can now run render tasks only once.
* Fixed an issue for the maya renderer related to the static inline variables.
* Re-enabled multi-threading.

### DIFF
--- a/src/settings.hpp
+++ b/src/settings.hpp
@@ -4,7 +4,7 @@ namespace wr::settings
 {
 
 	static const constexpr bool use_multithreading = true;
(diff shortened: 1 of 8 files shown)


src/settings.hpp
```diff
@@ -3,7 +3,7 @@
 namespace wr::settings
 {
 
-	static const constexpr bool use_multithreading = false;
+	static const constexpr bool use_multithreading = true;
```
LaisoEmilio: is this wanted?

PR #303, opened 2019-05-16
Feature uv scaling

DBouma:
* Scales the material's UV's of the following individually
- Albedo
- Normal
- Roughness
- Metallic
- Emissive
- AO 

SetConstants should be called from maya to set scales, Materials can be created by passing in the extra struct "MaterialUVScales".

Works on both deferred and raytracing.

### DIFF
--- a/src/structs.hpp
+++ b/src/structs.hpp
@@ -19,7 +19,7 @@ namespace wr
 
 	struct MaterialUVScales
 	{
-		float m_albedo_scale = 1.0f
+		float m_albedo_scale = 1.0f;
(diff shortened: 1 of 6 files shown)


src/structs.hpp
```diff
@@ -17,6 +17,16 @@ namespace wr
 		std::uint32_t m_id = 0;
 	};
 
+	struct MaterialUVScales
+	{
+		float m_albedo_scale = 1.f
```
tntmeijs: Consider using 1.0f instead of 1.f. The former seems to be used more often in the codebase.

PR #305, opened 2019-05-16
Replaced sentry with bugsplat

florianschut:
Replaced Sentry with Bugsplate since Sentry is quite useless and BugSplat allows us to download both the dumps and provides easy uploading and downloading of debug symbols for each release

### DIFF
--- a/tests/demo/demo.cpp
+++ b/tests/demo/demo.cpp
@@ -62,9 +62,7 @@ void startCrashpad()
 	base::FilePath handler(L"deps/crashpad/out/Release/crashpad_handler.exe");
 	// URL used to submit minidumps to
 	std::string url;
-	url = "https://";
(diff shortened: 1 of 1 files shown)


tests/demo/demo.cpp
```diff
@@ -61,9 +61,16 @@ void startCrashpad()
 	// Path to the out-of-process handler executable
 	base::FilePath handler(L"deps/crashpad/out/Release/crashpad_handler.exe");
 	// URL used to submit minidumps to
-	std::string url("https://sentry.io/api/1455776/minidump/?sentry_key=7735122410eb4b6bbaec9f1f6fd43aff");
+	std::string url;
+	url = "https://";
```
DBouma: Why separate the URL parts?

florianschut: That indeed is no longer really needed

PR #310, opened 2019-05-20
Feature RTAO settings

florianschut:
Added settings for RTAO which are adjustable through the engine interface using ImGui. Allows easy access to EPSILON/bias, Radius, Strenght and Sample count.

### DIFF
--- a/resources/shaders/ambient_occlusion.hlsl
+++ b/resources/shaders/ambient_occlusion.hlsl
@@ -82,7 +82,7 @@ void AORaygenEntry()
         ao_value -= (1.0f/float(sample_count)) * TraceAORay(0, wpos, getCosHemisphereSample(rand_seed, normal), radius, 0);
     }
 
-    output[DispatchRaysIndex().xy].x = ao_value * power;
+    output[DispatchRaysIndex().xy].x = ao_value / power;
 }
 
 [shader("closesthit")]
--- a/src/render_tasks/d3d12_rtao_task.hpp
+++ b/src/render_tasks/d3d12_rtao_task.hpp
@@ -16,13 +16,19 @@
 
 namespace wr
 {
-	namespace rtaoSettings
+	struct RTAOSettings
 	{
-		static float bias = 0.01f;
-		static float radius = 1.f;
-		static float power = 1.f;
-		static int sample_count = 8u;
-	}
+		struct Runtime
+		{
+			float bias = 0.01f;
+			float radius = 1.f;
+			float power = 1.f;
+			int sample_count = 8;
+		};
+
+		Runtime m_runtime;
+	};
+
 	struct RTAOData
 	{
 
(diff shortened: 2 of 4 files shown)


resources/shaders/ambient_occlusion.hlsl
```diff
@@ -79,7 +79,7 @@ void AORaygenEntry()
         ao_value -= (1.0f/float(sample_count)) * TraceAORay(0, wpos, getCosHemisphereSample(rand_seed, normal), radius, 0);
     }
 
-    output[DispatchRaysIndex().xy].x = ao_value * power;
+    output[DispatchRaysIndex().xy].x = ao_value / power;
```
florianschut: Higher powers used to decrease the strength of the occlusion since lower values meaning less light thus higher occlusion

src/render_tasks/d3d12_rtao_task.hpp
```diff
@@ -204,10 +211,10 @@ namespace wr
...
 				temp::RTAO_CBData cb_data;
 				cb_data.m_inv_vp = DirectX::XMMatrixInverse(nullptr, camera->m_view * camera->m_projection);
-				cb_data.bias = 0.01f;
-				cb_data.radius = 1.f;
-				cb_data.power = 1.f;
-				cb_data.sample_count = 8u;
+				cb_data.bias = rtaoSettings::bias;
+				cb_data.radius = rtaoSettings::radius;
+				cb_data.power = rtaoSettings::power;
+				cb_data.sample_count = static_cast<unsigned int>(rtaoSettings::sample_count);
```
florianschut: ImGui doesn't work with unsigned int so we save this as int and static_cast it before passing it to the CB

src/render_tasks/d3d12_rtao_task.hpp
```diff
@@ -16,6 +16,13 @@
 
 namespace wr
 {
+	namespace rtaoSettings
```
VZout: style guide.

src/render_tasks/d3d12_rtao_task.hpp
```diff
@@ -16,6 +16,13 @@
...
+	{
+		static float bias = 0.01f;
```
VZout: We have a proper way to deal with settings. See hbao or ansel.

src/render_tasks/d3d12_rtao_task.hpp
florianschut: Which style? Style guide is inconsistent and uses both snake and pascal case

florianschut: Namespace is removed entirely

src/render_tasks/d3d12_rtao_task.hpp
florianschut: Settings now use the proper system!

PR #311, opened 2019-05-20
Fix resolution scaling

DBouma:
* Fixed crash after resizing and swapping frame graphs. https://github.com/TeamWisp/WispRenderer/issues/300

* Increased window resizing performance.
* Window resize callbacks only work when the resolution actually changes to prevent excessive resize calls. 
* Set minimum resolution to prevent dragging the window into negative resolution.
* These changes also allowed me to finally run path tracer on fallback.


src/frame_graph/frame_graph.hpp
```diff
@@ -261,6 +261,14 @@ namespace wr
...
+			{
+				WaitForCompletion(i);
+				m_should_execute[i] = false;
+			}
+
+			render_system.WaitForAllPreviousWork();
+
 			for (decltype(m_num_tasks) i = 0; i < m_num_tasks; ++i)
 			{
 				WaitForCompletion(i);
```
VZout: this can now be renmoved.

tests/demo/demo.cpp
```diff
@@ -112,7 +120,11 @@ int WispEntry()
 	startCrashpad();
 
 	render_system = std::make_unique<wr::D3D12RenderSystem>();
-	auto window = std::make_unique<wr::Window>(GetModuleHandleA(nullptr), "D3D12 Test App", 1280, 720);
+
+	curRes.width = 1280;
```
VZout: Not very nice.

tests/demo/demo.cpp
```diff
@@ -172,6 +184,11 @@ int WispEntry()
 
 	window->SetResizeCallback([&](std::uint32_t width, std::uint32_t height)
 	{
+		if (curRes.width == width && curRes.height == height)
```
VZout: should probably be done by the render system.

src/d3d12/d3d12_renderer.cpp
```diff
@@ -273,6 +273,7 @@ namespace wr
 		d3d12::ResizeViewport(m_viewport, (int)width, (int)height);
 		if (m_render_window.has_value())
 		{
+			WaitForAllPreviousWork();
```
VZout: duplicate?

src/frame_graph/frame_graph.hpp
```diff
@@ -261,6 +261,14 @@ namespace wr
 		*/
 		inline void Resize(RenderSystem & render_system, std::uint32_t width, std::uint32_t height)
 		{
+			for (decltype(m_num_tasks) i = 0; i < m_num_tasks; ++i)
+			{
+				WaitForCompletion(i);
+				m_should_execute[i] = false;
+			}
+
+			render_system.WaitForAllPreviousWork();
```
VZout: duplicate?

src/frame_graph/frame_graph.hpp
```diff
@@ -261,6 +261,14 @@ namespace wr
 		*/
 		inline void Resize(RenderSystem & render_system, std::uint32_t width, std::uint32_t height)
 		{
+			for (decltype(m_num_tasks) i = 0; i < m_num_tasks; ++i)
+			{
+				WaitForCompletion(i);
+				m_should_execute[i] = false;
```
VZout: You should be able to remove this.

tests/demo/demo.cpp
DBouma: The reason why this is being done there is because first the render system is being resized and afterwards the framegraphs which are two separate calls from demo. Having it here prevents duplicate code.

PR #317, opened 2019-05-21
Feature log to file

sainsay:
Added directing all LOG messages to a file.
The user of the library has to create the file. see demo.cpp as example

Tested against. demo project and Wisp for Maya plug-in


sainsay:
Maya needs an addition to the plugin initialization to make use of it. But that change will not break anything if the "user" of wisp does not add the required lines in their project

src/util/logfile_handler.cpp
```diff
@@ -0,0 +1,47 @@
+#include "logfile_handler.hpp"
+
+#include <sstream>
+
+wr::LogfileHandler::LogfileHandler()
```
florianschut: You might want to consider giving `wr::LogfileHandler::LogfileHandler(std::filesystem::path& dir_path, std::string& file_name)` default argument values to ommit the argumentless constructor to avoid issues when updating on of the constructors

PR #318, opened 2019-05-23
Fix reduce warnings

DBouma:
* Reduced almost all warnings, there are still some [nodiscard] warnings and templating warnings that seemingly cant be fixed without breaking the project. 
* Changed order of multiplication to explicitly use proper format for scaling resolution (for safety)

### DIFF
--- a/src/d3d12/d3d12_resource_pool_texture.cpp
+++ b/src/d3d12/d3d12_resource_pool_texture.cpp
@@ -137,8 +137,6 @@ namespace wr
 		{
 			d3d12::CommandList* cmdlist = static_cast<d3d12::CommandList*>(cmd_list);
 
-			int frame = m_render_system.GetFrameIdx();
-
 			std::vector<d3d12::TextureResource*> unstaged_textures;
 			std::vector<d3d12::TextureResource*> need_mipmapping;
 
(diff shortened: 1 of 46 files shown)


src/window.hpp
```diff
@@ -82,8 +82,8 @@ namespace wr
 		HWND m_handle;
 		HINSTANCE m_instance;
 
-		uint32_t m_window_width = 0;
-		uint32_t m_window_height = 0;
+		int m_window_width = 0;
```
VZout: you should prefer int32 over int.

src/d3d12/d3d12_resource_pool_texture.cpp
```diff
@@ -482,18 +480,18 @@ namespace wr
...
+			mip_lvls = static_cast<std::uint32_t>(metadata.mipLevels);
 		}
 
 		Format texture_format = static_cast<wr::Format>(metadata.format);
 
 		d3d12::desc::TextureDesc desc;
 
-		desc.m_width = metadata.width;
-		desc.m_height = metadata.height;
+		desc.m_width = static_cast<std::uint32_t>(metadata.width);
```
VZout: All these static casts suppress the warning but a better solution would be to change the metadata.width type instead if possible.

Issue #320, opened 2019-05-24
Compute pipelines don't transition their render target.


VZout:
Currently the compute pipeline start and end functions don't transition the render target even though that would be expected by the user.

PR #321, opened 2019-05-24
Feature fg

VZout:
Opened #320. Noticed that while writing this commit but the fix wouldn't be contained to the frame graph.

* Fixed some issues with update settings.
* General clean up for the frame graph.

### DIFF
--- a/src/d3d12/d3d12_renderer.cpp
+++ b/src/d3d12/d3d12_renderer.cpp
@@ -231,7 +231,7 @@ namespace wr
 		scene_graph->Update();
 		scene_graph->Optimize();
 
-		frame_graph.Execute(*this, *scene_graph.get());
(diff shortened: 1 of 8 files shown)


src/d3d12/d3d12_renderer.cpp
```diff
@@ -462,23 +474,14 @@ namespace wr
 		{
 			LOGW("A render target has no transitions specified. Is this correct?");
 		}
-
-		d3d12::End(n_cmd_list);
 	}
 
 	void D3D12RenderSystem::StartComputeTask(CommandList * cmd_list, std::pair<RenderTarget*, RenderTargetProperties> render_target)
 	{
```
sainsay: functions StartComputeTask and StopComputeTask are now empty. concider using __declspec(deprecated("Not implemented")) or something similar

PR #322, opened 2019-05-24
Cleaner Try Catch, and a imgui sldier for the frustum far and near.

VZout:
Also ignoring /logs/ now.

### DIFF
--- a/src/frame_graph/frame_graph.hpp
+++ b/src/frame_graph/frame_graph.hpp
@@ -187,7 +187,6 @@ namespace wr
 
 			if constexpr (settings::use_multithreading)
 			{
-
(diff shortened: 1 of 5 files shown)


src/frame_graph/frame_graph.hpp
```diff
@@ -762,19 +761,17 @@ namespace wr
...
-					}
+					return std::any_cast<R>(m_settings[i].value());
 				}
 			}
 
 			LOGC("Failed to find task settings! Does your frame graph contain this task?");
 			return R();
 		}
+		catch (const std::bad_any_cast& e) {
+			LOGW("A task settings requested failed to cast to T. ({})", e.what());
```
sainsay: when settings cannot be found it logs a critical. if it is a bad cast it is only a warning. how severe is it when this operation fails?

Issue #278, opened 2019-05-08
`d3d12::CreateReadbackBuffer`'s descriptor is passed as a pointer.


VZout:
in line
```
[[nodiscard]] ReadbackBufferResource* CreateReadbackBuffer(Device* device, desc::ReadbackDesc* description);
```
`d3d12::ReadBackDesc` is passed as a pointer. This should probably be a const reference.

VZout:
Fixed in #323

Issue #276, opened 2019-05-08
`D3D12RenderSystem::Render` takes a shared pointer and a normal reference.


VZout:
Currently the `D3D12RenderSystem::Render` method takes a `std::shared_ptr<SceneGraph> const & scene_graph` and a `FrameGraph & frame_graph`. This is does not make sense. Both should just be references, pointers or smart pointers. Not 1 of each.

VZout:
Fixed in #323

Issue #117, opened 2019-01-11
Use C++'s "final" keyword to clarify design choice.


sainsay:
**Is your request related to a problem? Please describe.**
D3D12RenderSystem class can currently be inherited from but marks all virtual functions as final giving no use to the child class. maintaining D3D12RenderSystem when the RenderSystem class changes can introduce mistakes as it is not clear that D3D12RenderSystem is not supposed to be inherited from in current design philosophy.

**Describe the solution you'd like**
D3D12RenderSystem class marked with the final specifier and remove in necessary final specifiers from functions. this makes all future additions to RenderSystem when virtual to be final in the D3D12RenderSystem. 

**Describe alternatives you've considered**
none

**Additional context**
none

VZout:
Fixed in #323

PR #323, opened 2019-05-24
Refactored the readback buffer and `RenderSystem` slightly

VZout:
Fixes issue #278, #276  and #117

### DIFF
--- a/src/d3d12/d3d12_renderer.cpp
+++ b/src/d3d12/d3d12_renderer.cpp
@@ -539,14 +539,8 @@ namespace wr
 
 		d3d12::Begin(cmd_list, 0);
 
-		// Readback
-		d3d12::desc::ReadbackDesc readback_buffer_desc = {};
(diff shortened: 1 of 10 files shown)


src/d3d12/d3d12_renderer.cpp
```diff
@@ -145,7 +145,7 @@ namespace wr
 		m_texture_pools.push_back(CreateTexturePool());
 	}
 
-	CPUTextures D3D12RenderSystem::Render(std::shared_ptr<SceneGraph> const & scene_graph, FrameGraph & frame_graph)
+	CPUTextures D3D12RenderSystem::Render(SceneGraph& scene_graph, FrameGraph& frame_graph)
```
tntmeijs: Can't we make the frame_graph a const& as well?

VZout: We can't, the scene graph also can't. The renderer updates both depending if there are changes,

PR #325, opened 2019-05-24
Added orthographic camera

DBouma:
* Added orthographic camera
* Added TEMPORARY workaround for NaN pixels being output by deferred composition.
* Cleaned up some shader code to use explicit values.

### DIFF
--- a/src/scene_graph/camera_node.cpp
+++ b/src/scene_graph/camera_node.cpp
@@ -55,6 +55,16 @@ namespace wr
 		return std::pair<float, float>(m_projection_offset_x, m_projection_offset_y);
 	}
 
+	void CameraNode::SetOrthographicResolution(std::uint32_t width, std::uint32_t height)
+	{
+		m_ortho_res.m_width = static_cast<int>(width);
+
+		//Window resolution counts in negative on Y axis, therefore conversion to signed is needed.
+		m_ortho_res.m_height = static_cast<int>(height) * -1;
+
+		SignalChange();
+	}
+
 	void CameraNode::UpdateTemp(unsigned int frame_idx)
 	{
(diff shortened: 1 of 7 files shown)


src/scene_graph/camera_node.cpp
```diff
@@ -55,6 +55,16 @@ namespace wr
 		return std::pair<float, float>(m_projection_offset_x, m_projection_offset_y);
 	}
 
+	void CameraNode::SetOrthographicResolution(std::uint32_t width, std::uint32_t height)
+	{
+		m_ortho_res.m_width = static_cast<int>(width);
```
DBouma: The reason for this is because the window counts negative on screen, without this the Y coordinate will be flipped.

src/scene_graph/camera_node.hpp
```diff
@@ -33,6 +35,21 @@ namespace wr
 			float m_fov;
 		};
 
+		struct OrthographicResolution
+		{
+			explicit OrthographicResolution(OrthographicWidth width, OrthographicHeight height) : m_width(width), m_height(height)
+			{
+			}
+
+			/*OrthographicResolution(OrthographicWidth width, OrthographicHeight height) :
```
VZout: remove commented out code plz

src/scene_graph/camera_node.hpp
```diff
@@ -55,7 +72,8 @@ namespace wr
 			m_inverse_projection(),
 			m_view_projection(),
 			m_camera_cb(),
-			m_planes()
+			m_planes(),
+			m_ortho_res(OrthographicWidth(1280), OrthographicHeight(-720))
```
VZout: You don't really benefit from this approach, just use a std::pair with 2 integers for the width would be better. make sure to still default initialize though. Also the height shouldn't be -. Apply the minus later maybe?

PR #326, opened 2019-05-25
Switched the registries to C++20's designated initializers.

VZout:
* Registries now get destroyed properly.
* Removed the named type from the registries in favor of designated initialization.
* Now using std's `contains` function instead of my own.

### DIFF
--- a/src/engine_registry.cpp
+++ b/src/engine_registry.cpp
@@ -31,11 +31,11 @@ namespace wr
 		);
 
 	REGISTER(root_signatures::brdf_lut, RootSignatureRegistry)({
-		RootSignatureDescription::Parameters({
+		.m_parameters = {
 			ROOT_PARAM_DESC_TABLE(ranges_brdf, D3D12_SHADER_VISIBILITY_ALL)
-		}),
-		RootSignatureDescription::Samplers({ })
-		});
+		},
+		.m_samplers = { }
+	});
 
 
 	//Basic Deferred Pass Root Signature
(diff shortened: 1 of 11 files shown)


src/engine_registry.cpp
```diff
@@ -31,11 +31,11 @@ namespace wr
 		);
 
 	REGISTER(root_signatures::brdf_lut, RootSignatureRegistry)({
-		RootSignatureDescription::Parameters({
+		.m_parameters = {
```
DBouma: Not entirely sure what the . prefix does here, same goes for the other prefixes, it is explicit and clear though but might be confusing for the team

src/engine_registry.cpp
```diff
@@ -75,331 +75,329 @@ namespace wr
...
-		PipelineDescription::TopologyType(TopologyType::TRIANGLE)
-		}
-	);
+		.m_vertex_shader_handle = std::nullopt,
+		.m_pixel_shader_handle = std::nullopt,
+		.m_compute_shader_handle = shaders::brdf_lut_cs,
+		.m_root_signature_handle = root_signatures::brdf_lut,
+		.m_dsv_format = Format::UNKNOWN,
+		.m_rtv_formats = { Format::R16G16_FLOAT },
+		.m_num_rtv_formats = 1,
```
DBouma: m_num_rtv_formats is a uint, might be a good idea to explicitly assign an unsigned value. Not too important though, just consistancy as we do this for most of the project.

src/engine_registry.cpp
```diff
@@ -444,441 +442,441 @@ namespace wr
...
 
 	// Cone of confusion
 	REGISTER(shaders::dof_coc, ShaderRegistry) ({
-		ShaderDescription::Path("resources/shaders/dof_coc.hlsl"),
-			ShaderDescription::Entry("main_cs"),
-			ShaderDescription::Type(ShaderType::DIRECT_COMPUTE_SHADER)
-		});
+		.path = "resources/shaders/dof_coc.hlsl",
+		.entry = "main_cs",
+		.type = ShaderType::DIRECT_COMPUTE_SHADER
```
DBouma: Perhaps make this variable more explicit? Like ".shader_type", just a suggestion

src/pipeline_registry.hpp
```diff
@@ -16,37 +16,22 @@ namespace wr
...
-
-		InputLayout m_input_layout = InputLayout({});
+		std::optional<RegistryHandle> m_vertex_shader_handle;
+		std::optional<RegistryHandle> m_pixel_shader_handle;
+		std::optional<RegistryHandle> m_compute_shader_handle;
+		RegistryHandle m_root_signature_handle;
+
+		Format m_dsv_format;
+		std::array<Format, 8> m_rtv_formats;
+		unsigned int m_num_rtv_formats;
```
DBouma: Perhaps declare this as std::uint32_t for consistancy?

VZout:
Merging, its not perfect but nothing related to the changes itself. the system wasn't perfect to begin with.

PR #331, opened 2019-05-27
Feature rtao large scenes

florianschut:
Replace the depth buffer with a position buffer which severely reduces self-intersection. Also added sample-count dropoff over a distance which reduces the intensity as a result, this is intended behavior. I also fixed Power to be the exponent instead of it making the entire scene darker. This is now inline with HBAO+.
Also removed the Hit shader which is a theoretical improvement proposed by  [EA Seed](https://media.contentapi.ea.com/content/dam/ea/seed/presentations/2019-ray-tracing-gems-chapter-25-barre-brisebois-et-al.pdf#page=15&zoom=100,0,292). No noticeable results came out of some quick tests but should allow the drivers to better optimize the rays.


resources/shaders/ambient_occlusion.hlsl
```diff
@@ -3,7 +3,7 @@
 
 RWTexture2D<float4> output : register(u0); // x: AO value
 Texture2D gbuffer_normal : register(t1);
-Texture2D gbuffer_depth : register(t2);
+Texture2D gbuffer_position: register(t2);
```
VZout: What is the reason for this change? Also I want to see a performance comparison with deferred master and deferred your branch.

tests/demo/demo.cpp
```diff
@@ -21,12 +21,13 @@
...
 #include "scene_emibl.hpp"
 #include "scene_spheres.hpp"
+#include "scene_largescene.hpp"
 
 #include "model_loader_assimp.hpp"
 #include "model_loader_tinygltf.hpp"
 #include "d3d12/d3d12_dynamic_descriptor_heap.hpp"
 
-#define SCENE viknell_scene
+#define SCENE emibl_scene
```
VZout: revert plzy

tests/demo/scene_largescene.hpp
```diff
@@ -0,0 +1,80 @@
+
```
VZout: please remove this file. koen will have the same file and that one has lights placed on the sun temple candle thingies.

resources/shaders/ambient_occlusion.hlsl
VZout: with this change I mean using gbuffer_position instead of depth.

florianschut: Due to the non-linear nature of the depth buffer, it can't provide sufficient accuracy for RTAO. This was the primary reason for our insane self-intersection. I will test the performance of this change to be able to make a better-educated opinion.

LaisoEmilio: Using a linear depth buffer would be less bandwidth heavy, as the buffer you're using is an RGBA16 bit buffer, while a linear depth buffer would only require a single-channel buffer.
That said, I'm going to accept this change and write it down for the optimization phase, if time allows, as I would like you to be able to move to your next task.

tests/demo/demo.cpp
LaisoEmilio: revert, viktor is maaaad

LaisoEmilio: he jelous that emibl looks better than viknell.

LaisoEmilio: Seriously tho, revert. Emibl takes way too much to load

PR #334, opened 2019-05-27
Feature var roughness 3

Nielsbishere:
Variable roughness with 1/4 samples per pixel. Currently needs a denoiser (temporal accumulation & bilateral cleanup). If denoised will look rough and not wet anymore.


src/frame_graph/frame_graph.hpp
```diff
@@ -21,14 +21,44 @@
...
 
-#define EXPAND(x) x // Because msvc handles the preprocessor differently
-#define FG_DEPS(N, ...) EXPAND(FG_DEPS##N(__VA_ARGS__))
-#define FG_DEPS1(A) { typeid(A) }
-#define FG_DEPS2(A, B) { typeid(A), typeid(B) }
-#define FG_DEPS3(A, B, C) { typeid(A), typeid(B), typeid(C) }
-#define FG_DEPS4(A, B, C, D) { typeid(A), typeid(B), typeid(C), typeid(D) }
-#define FG_DEPS5(A, B, C, D, E) { typeid(A), typeid(B), typeid(C), typeid(D), typeid(E) }
-#define FG_DEPS6(A, B, C, D, E, F) { typeid(A), typeid(B), typeid(C), typeid(D), typeid(E), typeid(F) }
+namespace fg_deps
```
VZout: I choose macro's to reduce the initialization time. I don't mind changing it to template magic. But this approach is far to convoluted. This can be simplified a lot. Also I'm not a fan of the style used here. Might be easiest to revert these changes and make a new pull request with a better approach.

VZout: (I think it would be easier to revert since you need to go to all tasks again to change it. so might be good to move that to a different commit so I can merge this.)

src/d3d12/d3d12_functions.hpp
```diff
@@ -116,13 +116,13 @@ namespace wr::d3d12
...
+	void SetShaderSRV(wr::d3d12::CommandList* cmd_list, uint32_t rootParameterIndex, uint32_t descriptorOffset, d3d12::DescHeapCPUHandle& handle, uint32_t descriptorCount = 1);
 	void SetShaderUAV(wr::d3d12::CommandList* cmd_list, uint32_t rootParameterIndex, uint32_t descriptorOffset, TextureResource* tex);
-	void SetShaderUAV(wr::d3d12::CommandList* cmd_list, uint32_t rootParameterIndex, uint32_t descriptorOffset, d3d12::DescHeapCPUHandle& handle);
+	void SetShaderUAV(wr::d3d12::CommandList* cmd_list, uint32_t rootParameterIndex, uint32_t descriptorOffset, d3d12::DescHeapCPUHandle& handle, uint32_t descriptorCount = 1);
 	void SetRTShaderSRV(wr::d3d12::CommandList* cmd_list, uint32_t rootParameterIndex, uint32_t descriptorOffset, TextureResource* tex);
-	void SetRTShaderSRV(wr::d3d12::CommandList* cmd_list, uint32_t rootParameterIndex, uint32_t descriptorOffset, d3d12::DescHeapCPUHandle& handle);
+	void SetRTShaderSRV(wr::d3d12::CommandList* cmd_list, uint32_t rootParameterIndex, uint32_t descriptorOffset, d3d12::DescHeapCPUHandle& handle, uint32_t descriptorCount = 1);
```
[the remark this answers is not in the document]
Nielsbishere: 🤔 I added "descriptorCount" but the other ones except cmd_list use a different styleguide as well

src/frame_graph/frame_graph.hpp
Nielsbishere: I'll revert it and put it into a separate PR. There's not a lot of additional initialization time using this technique, if I make it inline it should be relatively fast; and it only initializes it once and then caches it. It doesn't require the user to specify how many arguments are used (avoiding user errors) and is infinitely expandable (if for some reason there would be more than 6 dependencies; which would probably never happen). 
```cpp
	static fg_deps::type deps = fg_deps::make<args...>();
	return deps;
```
^^ it is ran once and then just uses the vector created

Nielsbishere: https://github.com/TeamWisp/WispRenderer/issues/340 moved to an issue instead. It's removed now

VZout: Yes I get it only runs once. But the way you wrote that code is highly convoluted as well which I why I don't like it.

resources/shaders/rt_hybrid.hlsl
```diff
@@ -133,14 +137,47 @@ float3 unpack_position(float2 uv, float depth)
...
+		nextRand(rand_seed);
+		xi = hammersley2d(rand_seed, 8192);
+		H = importanceSamplePdf(xi, roughness, N, pdf);
+		L = reflect(-V, H);
+		NdotL = max(dot(N, L), 0);
+	}
+
+	float3 reflection = float3(0, 0, 0);
+	
+	if (NdotL >= 0)
```
LaisoEmilio: add scope please

resources/shaders/rt_hybrid.hlsl
```diff
@@ -259,7 +301,7 @@ void ReflectionHit(inout ReflectionHitInfo payload, in MyAttributes attr)
 	payload.cone = Propagate(payload.cone, 0, length(payload.origin - hit_pos));
 
 	// Calculate the texture LOD level 
-	float mip_level = ComputeTextureLOD(
+	float mip_level = 0; /*ComputeTextureLOD(			Mipmapping
```
LaisoEmilio: Commented out block of code? Leftover debug code or can it be deleted?

resources/shaders/spatial_reconstruction.hlsl
```diff
@@ -0,0 +1,204 @@
...
+	for (uint i = 0; i < sample_count; ++i)
+	{
+		//Get sample related data
+
+		float randomVar = nextRand(rand_seed);
+
+		const float2 neighbor_uv = sample_neighbor_uv(i, pix, uint2(width, height), randomVar);
+
+		const float depth_neighbor = depth_buffer.SampleLevel(nearest_sampler, neighbor_uv, 0).r;
+	    const float3 pos_neighbor = unpack_position(neighbor_uv, depth_neighbor);
```
LaisoEmilio: Wierd indentation

resources/shaders/spatial_reconstruction.hlsl
```diff
@@ -0,0 +1,204 @@
...
+
+		const float4 hitT = dir_hitT.SampleLevel(nearest_sampler, neighbor_uv, 0);
+        //const float3 hit_pos = hitT.xyz * hitT.w + pos_neighbor;
+
+        const float3 V_neighbor = normalize(camera_pos - pos_neighbor);
```
LaisoEmilio: Wierd indentation

resources/shaders/rt_hybrid.hlsl
Nielsbishere: It can't be deleted, it's there because of https://github.com/TeamWisp/WispRenderer/issues/284. Mipmapping does work with perfect reflections, but with variable roughness can sample from the incorrect mip at some angles. I don't want to remove it but I don't want to keep it in either.

resources/shaders/rt_reflection_shadows.hlsl
```diff
@@ -176,7 +220,9 @@ void HybridRaygenEntry()
 	{
 		// A value of 1 in the output buffer, means that there is shadow
 		// So, the far plane pixels are set to 0
-		output_refl_shadow[DispatchRaysIndex().xy] = float4(0, 0, 0, 0);
+		output_reflection[screen_co] = float4(0, 0, 0, 0);
+		output_reflection[screen_co] = 0;
```
VZout: 🌊

resources/shaders/pbr_util.hlsl
```diff
@@ -164,3 +164,45 @@ float2 SampleSphericalMap(float3 v)
 	uv += 0.5f;
 	return uv;
 }
+
+// Brian Karis, Epic Games "Real Shading in Unreal Engine 4"
+// Modified version to do pdf and tangent to world conversions
+float3 importanceSamplePdf(float2 xi, float a, float3 N, inout float pdf) {
```
VZout: Rename this to `importanceSample_GGX_PDF` please and take over the implementation from that one so if we change something in one we can easily change it in the other.

resources/shaders/lighting.hlsl
```diff
@@ -127,7 +127,7 @@ float3 shade_pixel(float3 pos, float3 V, float3 albedo, float metallic, float ro
 	return res + emissive;
 }
 
-float4 DoShadowAllLights(float3 wpos, float3 V, float3 normal, float metallic, float roughness, float3 albedo, uint depth, inout float rand_seed)
+float4 DoShadowAllLights(float3 wpos, float3 V, float3 normal, float metallic, float roughness, uint depth, inout float rand_seed)
```
VZout: Why? doesn't seem used.

PR #343, opened 2019-05-28
Added support for removal of skybox nodes.

LaisoEmilio:
The render system has now a default 2x2 cubemap to use in the case of absence of a skybox. This is due to the descriptor heap implementation we are using, as we need to have something bound to the pipeline in those slots. This will be expanded further to support plain colors for skyboxes in a separate brach.


src/d3d12/d3d12_renderer.cpp
```diff
@@ -1348,4 +1349,9 @@ namespace wr
 		}
 	}
 
+	void D3D12RenderSystem::CreateDefaultResources()
```
VZout: the default textures should be moved in here probably.

src/d3d12/d3d12_renderer.cpp
```diff
@@ -1348,4 +1349,9 @@ namespace wr
 		}
 	}
 
+	void D3D12RenderSystem::CreateDefaultResources()
+	{
+		m_default_cubemap = m_texture_pools.at(0)->CreateCubemap("DefaultResource_Cubemap", 2, 2, 1, wr::Format::R8G8B8A8_UNORM, false);
```
VZout: Missing extension or something?

src/render_tasks/d3d12_deferred_composition.cpp
```diff
@@ -218,12 +218,18 @@ namespace wr
...
 				//GetSkybox
 				auto skybox = scene_graph.GetCurrentSkybox();
-				if (skybox != nullptr)
+				if (skybox)
 				{
 					//data.out_skybox = static_cast<wr::d3d12::TextureResource*>(pred_data.in_radiance.m_pool->GetTexture(pred_data.in_radiance.m_id));
 					data.out_skybox = static_cast<wr::d3d12::TextureResource*>(skybox->m_skybox->m_pool->GetTextureResource(skybox->m_skybox.value()));
 					d3d12::CreateSRVFromTexture(data.out_skybox);
 				}
+				else
```
VZout: This is turning into really shitty duplicate code. We should probably make a function that returns the texture for us instead.

src/d3d12/d3d12_renderer.cpp
LaisoEmilio: No, that is just a name for the resource itself, useful for debugging

src/render_tasks/d3d12_deferred_composition.cpp
LaisoEmilio: Reworked everythig as discussed privately

src/d3d12/d3d12_renderer.cpp
LaisoEmilio: moved the default textures out of texture pool and gave it to the renderer

src/scene_graph/scene_graph.hpp
```diff
@@ -128,6 +128,8 @@ namespace wr
 		std::vector<std::shared_ptr<LightNode>> m_light_nodes;
 		std::vector< std::shared_ptr<SkyboxNode>> m_skybox_nodes;
 
+		std::shared_ptr<SkyboxNode> m_default_skybox = nullptr;
```
VZout: the default skybox should probably be part of the scene graph tree itself for consistent behavior.

LaisoEmilio: Default skybox is a "fallback" for when a user deletes the last skybox in the scene tree. Adding it to the tree would allow its deletion which is not expected behaviour. Due to the descriptor heap system we're using, something NEEDS to be bound in the skybox slots even if the shaders are not going to sample from it, the default skybox has a 2x2 cubemap that is used as a fallback in case there's not skybox nodes available at render time

src/util/defines.hpp
```diff
@@ -88,4 +88,9 @@ decltype(wr::SceneGraph::m_update_lights_func_impl) wr::SceneGraph::m_update_lig
 decltype(wr::SceneGraph::m_update_transforms_func_impl) wr::SceneGraph::m_update_transforms_func_impl = [](wr::RenderSystem* render_system, wr::SceneGraph& scene_graph, std::shared_ptr<wr::Node>& node) \
 { \
 	static_cast<renderer_type*>(render_system)->function(scene_graph, node); \
+};
+#define LINK_SG_DELETE_SKYBOX(renderer_type, function) \
```
VZout: look at this github highlighting though :o

src/d3d12/d3d12_renderer.cpp
```diff
@@ -950,6 +952,20 @@ namespace wr
 
 	}
 
+	void D3D12RenderSystem::Delete_Skybox(SceneGraph& scene_graph, std::shared_ptr<SkyboxNode>& skybox_node)
```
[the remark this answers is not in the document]
LaisoEmilio: Just github, it's after the updates

src/util/defines.hpp
LaisoEmilio: :o

PR #344, opened 2019-05-28
Feature improved bloom

DBouma:
* Improved bloom algorithm to approximate a much larger kernel without the cost of such a kernel.
* No longer blooms skybox
* Made sure the bright output includes emissive materials.
* Cleaned up shader code, all bloom blurring uses the same shader now.
* Improved DoF to reduce dark outlines on in focus objects.


src/render_tasks/d3d12_bloom_eighth_res.hpp
```diff
@@ -0,0 +1,169 @@
+#pragma once
```
DBouma: bloom_half , quarter, eighth and sixteenth tasks are all duplicate tasks with slight modifications to resolution and blurring factors.

LaisoEmilio: Consider what we talked about in the lab with Viktor about this approach. Please create an issue in github so we don't forget to optimize this in the future.

PR #346, opened 2019-05-28
The scene graph now has the option to cull models far from the camera (RT)

Nielsbishere:
Added AABB->Sphere intersection, added shortcuts to get AABB data. Fixed AABB struct size (was too big). Added const and references when necessary


Nielsbishere:
https://github.com/TeamWisp/WispRenderer/issues/347

PR #355, opened 2019-06-04
Add first batch of Graphics Settings

florianschut:
Includes imgui_graphics_settings which controls the settings from Ansel, HBAO, Acceleration structure and reloading. 
Settings are also added to cubemap convolution but the usage of these is postponed due to issues and priorities.
Also adds Ansel to the hybrid pipeline.
The settings only show for the currently active render tasks


src/imgui_graphics_settings.hpp
```diff
@@ -0,0 +1,97 @@
...
+			ImGui::LabelText("", "RTAO");
+			ImGui::Separator();
+			
+				ImGui::DragFloat("Bias", &rtao_user_settings.m_runtime.bias, 0.01f, 0.0f, 100.f);
+				ImGui::DragFloat("Radius", &rtao_user_settings.m_runtime.radius, 0.1f, 0.0f, 1000.f);
+				ImGui::DragFloat("Power", &rtao_user_settings.m_runtime.power, 0.1f, 0.0f, 10.f);
+				ImGui::DragInt("SPP", &rtao_user_settings.m_runtime.sample_count, 1, 0, 1073741824);
+
+					frame_graph->UpdateSettings<wr::RTAOData>(rtao_user_settings);
+			ImGui::Dummy(ImVec2(0.0f, 10.0f));
```
MeineZ: Indenting

src/imgui_graphics_settings.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+	void GraphicsSettings(wr::FrameGraph* frame_graph)
+	{
+		if (graphics_settings_open)
+		{
+			ImGui::Begin("Graphics Settings", &graphics_settings_open);
+
+			if (frame_graph->HasTask<wr::RTAOData>())
+			{
+			ImGui::LabelText("", "RTAO");
```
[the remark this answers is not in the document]
florianschut: Fixed on newer branch

src/imgui_graphics_settings.hpp
florianschut: Fixed on newer branch

src/imgui_graphics_settings.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+
+	static wr::RTAOSettings rtao_user_settings;
+	static wr::HBAOSettings hbao_user_settings;
+	static wr::AnselSettings ansel_user_settings;
+	static wr::ASBuildSettings as_build_user_settings;
+
+	void GraphicsSettings(wr::FrameGraph* frame_graph)
+	{
+		if (graphics_settings_open)
```
VZout: I prefer 1 window for every task. That would be easier to order when using the demo.
you could just close the tasks you don't use.

florianschut: done

florianschut:
Fake news! Radius is not set to 0.0f

PR #359, opened 2019-06-06
[FIX] Fixed meshes randomly starting as hidden.

161563:
Fixed meshes randomly starting as hidden when running on release, caused by the variable not being initialized properly, thus initializing to a random value on release.


florianschut:
Tutti belle!

PR #368, opened 2019-06-12
Throw on texture load fail

MeineZ:
In WispForMaya, we have to catch when a texture failed to load. A LOGC triggers a debug break which WispForMaya can't catch. I changed this to LOGE (so it still outputs the error) and it now returns an invalid texture handle. So, whenever a texture failed to load, WispForMaya can catch it and handle it accordingly.

### DIFF
--- a/src/d3d12/d3d12_resource_pool_texture.hpp
+++ b/src/d3d12/d3d12_resource_pool_texture.hpp
@@ -61,9 +61,14 @@ namespace wr
 
 		d3d12::TextureResource* GetTextureResource(TextureHandle handle) final;
 
-		// Load a texture from file.
-		// Returns the texture handle to the loaded texture.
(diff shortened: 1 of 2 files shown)


src/d3d12/d3d12_resource_pool_texture.hpp
```diff
@@ -60,6 +60,10 @@ namespace wr
 		void ReleaseTemporaryResources() final;
 
 		d3d12::TextureResource* GetTextureResource(TextureHandle handle) final;
+
+		// Load a texture from file.
```
VZout: Not following the style guide. doxygen won't generate proper docs from this.

LaisoEmilio: This

PR #369, opened 2019-06-12
Added a versioning system for wisp.

VZout:
Added a versioning system that is fully constexpr and easy to increment.


src/version.hpp
```diff
@@ -0,0 +1,97 @@
+/*!
+A constexpr way to obtain program version information.
+Copyright © 2018, Viktor Zoutman
```
tntmeijs: ... and team? 😭

VZout: :| I wrote this code a years ago forgot to update the license.

src/version.hpp
```diff
@@ -0,0 +1,97 @@
...
+#include <cstdint>
+#include <cstdio>
+#include <stdexcept>
+
+namespace wr
+{
+
+	struct Version
+	{
+		std::uint32_t m_major = 0, m_minor = 0, m_patch = 0;
```
LaisoEmilio: I still don't like this, even if our code standard doesn't mention it :P your call though :)

src/version.hpp
```diff
@@ -0,0 +1,97 @@
...
+	};
+
+	namespace std_cexpr
+	{
+		constexpr size_t strlen(const char* str) {
```
LaisoEmilio: curly brackets in new line please :)

PR #378, opened 2019-06-15
Fix merge bloom passes

DBouma:
* Reduced number of render tasks by 6 
* Increased performance by 0.6~0.75 ms on GTX 980m. 
* Solved some previously unnoticed bloom bugs.
* Increased quality of bloom.


resources/shaders/pp_bloom_extract_bright.hlsl
```diff
@@ -26,12 +26,19 @@ void main_cs(int3 dispatch_thread_id : SV_DispatchThreadID)
 
 	float brightness = dot(final_color, float3(0.2126f, 0.7152f, 0.0722f));
 
-	if (brightness > 1.0f && g_depth.SampleLevel(s1, uv, 0).r < 1)
+	for (int i = -1; i < 2; ++i)
```
VZout: seems a bit hardcoded.

DBouma: Yes this is meant to stabilize the extraction pass, I don't want to do a bilateral filter to stabalize the extraction as this would be too expensive, a small diagonal offset makes a big difference.

PR #384, opened 2019-06-17
Feature headless

Nielsbishere:
Fixed headless rendering; you now create a window without setting the Windows API instance. This won't create a OS-backed window but rather a virtual one.


VZout:
Still uses triple buffering. This is not a good enough solution.

Nielsbishere:
Why not? @VZout using triple buffering allows the GPU to keep on working on the next frames while you readback from the result from a while ago

PR #388, opened 2019-06-18
Added a Qt-style copyright notice to the top of every source file

tntmeijs:
I've added a Qt-style copyright comment to the top of every source file.
It is important that every team member adds this header to new source files.

The project license has also been updated to use the Apache 2.0 license instead of the old one. This is more permissive and encourages others to contribute to the project.


tntmeijs:
There are some issues with hidden characters. Don't approve this just yet.

tntmeijs:
It should work now!

Issue #387, opened 2019-06-18
Memory leaks


Nielsbishere:
**Is your request related to a problem? Please describe.**
There's several memory leaks that result in problems when the maya plugin unloads wips.
This thread will be used to note the findings of the inspection.

**Describe the solution you'd like**
These instances should be fixed.

**Describe alternatives you've considered**
N.A.

**Additional context**
Create an exit callback (call onexit(myCallback)), place a breakpoint on start and in the callback and take 2 snapshots to diff.

Nielsbishere:
```cpp
// Make sure we free the data objects we allocated.
for (auto& data : m_data)
{
	delete data;
}
```
__In frame_graph.hpp Destroy function__
Auto doesn't provide clarity about the type; but the type is void*. This means that it doesn't call free or the destructor of the object.

Nielsbishere:
PhysicsEngine::destroy is never being called, leaving behind bullet's resources.
## Memory impact
The biggest impact as bullet allocates several memory pools (est. 60%)

Nielsbishere:
texture_pool ptr is potentially never reset; leaving behind some memory.
## Memory impact
Not that much (est. 5%)

Nielsbishere:
- [ ] aiScene* isn't being cleaned up
- [ ] Viknell probably still has materials left over
- [ ] Sponza probably still has materials left over

Nielsbishere:
+ [ ] Models don't clean up their allocated materials
+ [ ] Physics component doesn't clean up the rigid bodies, shapes, etc.

Nielsbishere:
+ [ ] Thread pool enqueue leaks memory

Nielsbishere:
+ [ ] Build acceleration structure task doesn't clean up the CB

Nielsbishere:
Possible issues to look into further:
Left-over data includes; C++ initialization data which should be taken care of by the program finishing unloading. Other than that; file streams seem to have temporary memory that may or may not be deleted after program end.

Nielsbishere:
Even though CPU memory leaks are mostly cleaned up, there could potentially still be GPU memory leaks.

PR #396, opened 2019-06-21
Feature fix memleak 2

Nielsbishere:
#387


Nielsbishere:
https://github.com/TeamWisp/WispRenderer/pull/390 but without git's problematic diffing.

PR #405, opened 2019-07-02
Improved Hybrid renderer performance.

DBouma:


resources/shaders/dxr_global.hlsl
```diff
@@ -24,7 +24,7 @@
 #define SOFT_SHADOWS
 #define MAX_SHADOW_SAMPLES 1
 
-//#define PERFECT_MIRROR_REFLECTIONS
+#define PERFECT_MIRROR_REFLECTIONS
```
Nielsbishere: No.

resources/shaders/dxr_reflection_functions.hlsl
```diff
@@ -85,8 +85,9 @@ float4 DoReflection(float3 wpos, float3 V, float3 N, uint rand_seed, uint depth,
 
 		// Shoot perfect mirror ray if enabled or if it's a recursion or it's almost a perfect mirror
 
-		if (depth > 0 || roughness < 0.05f)
+		if (depth > 0 || roughness < 0.075f)
```
Nielsbishere: Why decrease the roughness value required to get a perfect mirror?

resources/shaders/dxr_reflection_functions.hlsl
```diff
@@ -165,9 +166,5 @@ float4 DoReflection(float3 wpos, float3 V, float3 N, uint rand_seed, uint depth,
 
 		#endif
 
-	#else
```
Nielsbishere: Why remove this? It allows perfect mirrors.

Nielsbishere (changes requested):

Nielsbishere:
This doesn't even solve infinite recursions.

resources/shaders/dxr_global.hlsl
161563: Doing this should disable rough reflections. PERFECT_MIRROR_REFLECTIONS shouldn't normally be defined.

resources/shaders/dxr_reflection_functions.hlsl
```diff
@@ -85,8 +85,9 @@ float4 DoReflection(float3 wpos, float3 V, float3 N, uint rand_seed, uint depth,
 
 		// Shoot perfect mirror ray if enabled or if it's a recursion or it's almost a perfect mirror
 
-		if (depth > 0 || roughness < 0.05f)
+		if (depth > 0 || roughness < 0.075f)
 			return float4(TraceReflectionRay(wpos, N, reflected, rand_seed, depth, cone, dir_t), 1);
+	#endif
```
161563: This means that turning on PERFECT_MIRROR_REFLECTIONS will instead give you rough reflections, including for recursions, which means the define now does the opposite of what it should do.

161563 (changes requested):
How does this fix stop infinite recursions from happening? None of these changes suggest that they fixed it, or even that we've had them in the first place.

resources/shaders/dxr_global.hlsl
DBouma: it does not.

